PR #2, opened 2021-12-17
Use Gazelle to generate docs

shs96c:
Skylib now contains a language plugin that generates `bzl_library` targets for you. The template rules package uses this, so might as well hook it up here too.

By doing this, the top-level build file is now unusable by other libraries, since it will add a requirement to download gazelle and rules go, and that's pretty heavy, so I have no intention of adding them to `contrib_rules_jvm_deps`. This means the default PMD config needs to be moved, but the new location makes more sense anyway.


Yannic (commented):
LGTM in general.

The PMD changes seem unrelated (although might be prerequisite), so maybe split them out? Would also be nice to have one "add plumbing" PR that adds gazelle and one "update build files" PR that applies it.

On CI, there should be a step that verifies running gazelle produces no diff.

shs96c:
The PMD move is related --- the introduction of gazelle added a heavyweight dependency (the Go toolchains) that most users of this ruleset won't use. By moving the PMD definitions into `//java` we continue to make that available, but also avoid the need for anyone to require anything from the top-level build file in regular use.

A previous PR pulled in the changes from the template project. If that has the gazelle check wired up, we're good. If not, I'll add it.

shs96c:
Thank you for the review, @Yannic!

PR #8, opened 2022-01-07
Disable the PMD analysis cache

shs96c:
We never set this in the first place, but users would see a
warning that they could improve build times by using it. Since
each test run is hermetic, we can't realistically do that, so
instead we disable the warning.

### DIFF
--- a/java/private/pmd.bzl
+++ b/java/private/pmd.bzl
@@ -5,6 +5,10 @@ def _pmd_test_impl(ctx):
         ctx.executable._pmd.short_path,
     ]
 
+    # We want to disable the suggestion to use the analysis cache
+    # https://pmd.github.io/latest/pmd_userdocs_incremental_analysis.html#disabling-incremental-analysis
+    cmd.extend(["-no-cache"])
+
     inputs = []
     transitive_inputs = depset()
 


thesayyn (approved):
Just an idea: would it speed up the compilation if you were to declare the cache analysis file and use it in subsequent builds? so that pmd skips unnecessary file analysis for unchanged files? 
as per the doc, it says it will check the checksum of the files so mtime is not going to be an issue here.

shs96c:
Given how bazel should be executing this in a sandbox without access to the wider filesystem, the cache it builds would be for the files seen in this one run of the action, and would then be discarded. It would be very surprising if subsequent builds could alter this file (and a gross violation of build hermeticity) Which is a long way of saying that I don't think the cache analysis file is really helpful for us :)

thesayyn:
> Given how bazel should be executing this in a sandbox without access to the wider filesystem, the cache it builds would be for the files seen in this one run of the action, and would then be discarded. It would be very surprising if subsequent builds could alter this file (and a gross violation of build hermeticity) Which is a long way of saying that I don't think the cache analysis file is really helpful for us :)

It seemed to me like `--cache` can point to a file that resides inside output-dir instead of a global cache location. we have done something similar in rules_nodejs here: https://github.com/bazelbuild/rules_nodejs/blob/89fbec3801487b522ac36f32a7a6b00ebf069695/packages/typescript/internal/ts_project.bzl#L186
Anyways I am not familiar with how pmd works internally so no gonna make too much noise :)

PR #11, opened 2022-01-17
Allow the checkstyle binary to be configured

### LINKED ISSUE #10: Checkstyle version should be configurable
Checkstyle as a project seems to rename and deprecate config settings relatively quickly, when fixing #3 I discovered that the sample configs already checked into this repo aren't even compatible with the Checkstyle version pinned in contrib_rules_jvm_deps.zip.

It's inconvenient for new users of this rule to have to manually update existing config files that might be quite old, and going forwards when Checkstyle makes config changes in new versions, either this project will have to stay stuck on the old version to prevent breaking downstream users, or users will be stuck on old releases until they can update their configs.

That being said this is probably a pretty low priority item right now, but might be worth considering in the future especially when a new major version of Checkstyle is released.

shs96c:
This allows users to select the version of checkstyle that they
want to use without being rail-roaded into taking the version
that ships with `rules_contrib_jvm`.

We also remove the `config_file` option, and instead always make
users specify the `checkstyle_config` that they want to use. This
will likely reduce boilerplate and duplication in people's build
files, and also makes it easier for us to figure out the right
binary to use.

Closes #10

### DIFF
--- a/java/private/checkstyle_config.bzl
+++ b/java/private/checkstyle_config.bzl
@@ -1,5 +1,6 @@
 CheckStyleInfo = provider(
     fields = {
+        "checkstyle": "The checkstyle binary to use.",
         "config_file": "The config file to use.",
         "output_format": "Output Format can be plain or xml.",
     },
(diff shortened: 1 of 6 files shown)


java/private/checkstyle_config.bzl
```diff
@@ -36,6 +39,15 @@ checkstyle_config = rule(
             doc = "Additional files to make available to Checkstyle such as any included XML files",
             allow_files = True,
         ),
+        "checkstyle_binary": attr.label(
+            doc = "Checkstyle binary to use.",
+            default = "@contrib_rules_jvm//java:checkstyle_cli",
+            executable = True,
```
thirtyseven, line 45: Maybe consider adding either a macro that takes a checkstyle jar and automatically defines the `java_binary` or add an example to the docs that shows how to define it. I don't know if end users would be expected to know the location of checkstyle's main class, and it seems to be pretty stable across versions.

thirtyseven (approved):

java/private/checkstyle_config.bzl
shs96c: Why not both? :)

shs96c: Done

thirtyseven (approved):

alexeagle (approved):

PR #12, opened 2022-01-19
Allow the spotbugs binary to be selected by the user

shs96c:
But continue to default to the current latest version of spotbugs
because that's just good manners.

This change mirrors the same one made for `checkstyle_test`, and
so we also include a macro for creating the target for the
spotbugs CLI tool.


java/private/spotbugs_config.bzl
```diff
@@ -1,17 +1,70 @@
+
+def spotbugs_binary(
```
blorente, line 2: Could we expose this through `//java:defs.bzl`?

shs96c: Yes!

alexeagle (approved):

PR #15, opened 2022-04-04
PMD: replace deprecated -filelist with --file-list

rdesgroppes:
This fixes a quite noisy recurring warning:
```
[date and time] net.sourceforge.pmd.PMD runPmd
WARNING: Some deprecated options were used on the command-line, including -filelist
[date and time] net.sourceforge.pmd.PMD runPmd
WARNING: Consider replacing it with --file-list
```

### DIFF
--- a/java/private/pmd.bzl
+++ b/java/private/pmd.bzl
@@ -19,7 +19,7 @@ def _pmd_test_impl(ctx):
         file_list,
         ",".join([src.path for src in ctx.files.srcs]),
     )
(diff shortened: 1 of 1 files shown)


illicitonion (approved):
Thanks for the contribution! This LGTM - one thing we should be aware of is it sets the minimum version of PMD at 6.41.0 which was released in November 2021, but that seems like a reasonable bump (and is the version pinned in this repo by default).

PR #17, opened 2022-04-20
Update the freeze-deps.py to process compat-enabled repositories

tjoneslo:
Add the processing to copy the compat-repositories code generated by the `generate_compat_repositories = True,` into the frozen deps zip file to allow correct processing.

Add documentation on how to use the freeze deps process in your code.


shs96c (approved):
I always forget about the compat libs stuff. Thank you for this!

PR #24, opened 2022-05-16
Only set --output_user_root in tests

illicitonion:
We were accidentally unconditionally setting this, often to a directory
which couldn't be properly written to.

Instead of looking for the configured test tmpdir, and falling back to
the global tmpdir if not set, only do this overriding _if_ the test
tmpdir is actually configured.


shs96c (approved):
LGTM. Thanks for fixing this!

PR #25, opened 2022-05-16
Stringify labels in error

illicitonion:
Currently the %#v printing of a Label is pretty hard to read - instead,
stringify them so they read cleanly as a user would expect.


shs96c (approved):
LGTM.

PR #26, opened 2022-05-16
Narrow filtering of imports in modules

illicitonion:
This ensures that when you have:

module/com/example/Foo
otherpackage/com/example/not_in_module/Bar

module's dependency on Bar isn't filtered out.


shs96c (approved):
This seems to make sense, and the tests are green, so LGTM

PR #28, opened 2022-05-20
Add two missing archive dependencies used by external system which 

tjoneslo:
are pulled in implicitly. But if used externally, this causes build failures when they are not included correctly.

Move some of the Gazelle dependencies and setup to separate files/packages

To avoid loading the enormous Go toolchain if not really using the Gazelle setup.


shs96c (approved):
Nice. Thank you for doing this!

PR #35, opened 2022-06-13
Use getQualifiedName

illicitonion:
There's a bug in javaparser where annotation class names looked up by
reflective detection have an additional . added to them which causes a
crash.

While we wait for the fix to be PR'd/merged/released upstream, use the
upstream method for getting a fully qualified name, rather than
splitting and joining.

Fixes https://github.com/bazel-contrib/rules_jvm/pull/35


shs96c (approved):
LGTM! Thank you!

PR #37, opened 2022-06-14
Skip JUnit 5 tests when assumptions are not met

bonigarcia:
I realized that when using the `junit5` runner in the `java_test_suite`, Jupiter tests fails when assumptions are not met. Instead, they should skipped.

I reproduced it with the test called `shouldBeSkipped` in class `AssumptionsJUnit5Test` (contained in this PR). This test, without the changes made on `CommandLineSummary` and `TestResult`, fails as follows:

Screenshot from 2022-06-14 12-58-32

With the new changes, the test is skipped, as expected:

Screenshot from 2022-06-14 15-00-00


shs96c (approved):
LGTM. Thank you for spotting this and providing a PR!

PR #40, opened 2022-06-22
gazelle: Fix deps ordering when using #keep

stevebarrau:
Fixes an ordering issue with using `#keep` comments in deps attribute.

The [mergeList](https://github.com/bazelbuild/bazel-gazelle/blob/a98769c9a26e312ac40d1f5b14f574fdcd4aa21a/rule/merge.go#L144) function expects both arguments to be lists of the same kind of `build.Expr`. The `dst` "deps" attribute is parsed as a list of `build.StringExpr`.

### DIFF
--- a/java/gazelle/resolve.go
+++ b/java/gazelle/resolve.go
@@ -3,7 +3,6 @@ package gazelle
 import (
 	"fmt"
 	"sort"
-	"strconv"
 
 	"github.com/bazel-contrib/rules_jvm/java/gazelle/private/java"
 	"github.com/bazelbuild/bazel-gazelle/config"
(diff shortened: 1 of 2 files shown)


shs96c (commented):
Should we have a test for this? Or is it the case that now this is fixed, it won't regress?

stevebarrau:
We do not have the class of test this change needs. The issue is not in the generate or the resolve. It's in the type used in the "deps" attribute by resolve. I am not sure it cannot regress, but I adding this test class would require a bit of work.

shs96c (approved):
Thanks for adding the test! LGTM

PR #44, opened 2022-06-29
Allow setting visibility for java_test_suite libraries

illicitonion:
Currently, there is no way for other packages to depend on the
implicitly generated test library. This allows for a BUILD file author
to opt in to making the generated test library more visible.

This re-uses the visibility attribute which is also currently used for
the visibility of the generated `test_suite`. This feels less fiddly
than introducing a separate attribute.


shs96c (approved):
Seems obvious in hindsight! Thank you!

PR #45, opened 2022-06-30
Bump io_grpc_grpc_java dep

illicitonion:
Recent protobuf appears to have introduced a change which breaks this
version, removing `google/protobuf/compiler/java/java_names.h`.

Also, we're already using 1.47.0 in `contrib_rules_jvm_deps`, so this is
consistent across the file.


shs96c (approved):
LGTM. Thank you for the update!

PR #49, opened 2022-07-05
Use rules_jvm_external install for package index

illicitonion:
https://github.com/bazelbuild/rules_jvm_external/pull/716 started
indexing jar files and adding the package index to the
maven_install.json.

This re-uses that data, rather than re-implementing parsing ourselves.
This has the advantage of no longer needing to check in and manage the
lifecycle of an additional cache file, and deletes a bunch of code.


third_party/go/repositories.bzl
```diff
@@ -55,12 +55,7 @@ def go_deps():
         sum = "h1:8/MDe6zCE7nYKXhlbamM1izUgJjw2ce7M5kd4QKF/3Y=",
         version = "v0.31.0",
     )
-    go_repository(
-        name = "com_github_bazelbuild_tools_jvm_autodeps",
-        importpath = "github.com/bazelbuild/tools_jvm_autodeps",
-        sum = "h1:wcw0i+MQc/Yo8RgkS09xSujJnOMCzZXD6LUxhKxGhMg=",
-        version = "v0.0.0-20180917073602-62694dd50b91",
-    )
+
```
stevebarrau, line 58: This is a gazelle limitation, until https://github.com/bazelbuild/bazel-gazelle/issues/999 is resolve, we usually use @EdSchouten solution:

```diff
# tools/update-dependencies.sh
  go mod tidy
- bazel run //:gazelle_go -- update-repos \
-     -from_file=go.mod \
-     -prune \
-     -to_macro "third_party/go/repositories.bzl%go_deps"
+ GO_DEPS_FILE="third_party/go/repositories.bzl"
+ bazel run //:gazelle -- update-repos -from_file=go.mod -prune -to_macro "${GO_DEPS_FILE}%go_deps"
+ sed '/^$/d' "$GO_DEPS_FILE" >"${GO_DEPS_FILE}.new"
+ mv "${GO_DEPS_FILE}.new" "$GO_DEPS_FILE"
+ bazel run //:buildifier
```

If we do update the tools/update-dependencies.sh script, we could:

1. move Java pin first given unpinned deps would fail the build of //:gazelle_go
2. drop `bazel fetch` of maven deps given we do not need the JARs on disk for parsejar anymore

stevebarrau (approved):
Thanks for cleaning this up @illicitonion ✨

third_party/go/repositories.bzl
illicitonion: Adopted - thanks for the pointer and context!

shs96c (approved):
Rubber stamping. Perhaps we should note that you need a specific version of `rules_jvm_external` for this to work?

illicitonion:
There's a note in the readme :) https://github.com/bazel-contrib/rules_jvm/pull/49/files#diff-7523dd5c8e9927a3bf70fe2cb40381b71314610c5d406372edd22df777b1c35cR87-R88

PR #50, opened 2022-07-05
Flesh out tests for java_test and java_test_suite generation

illicitonion:
We now have integration test cover for all four quadrants of the matrix of {module,per-package-BUILD-file} x {per-file-test,per-package suite}

We don't have great unit test coverage. More to come.

There are also several TODOs for stuff that looks probably wrong in the golden files, which I will clean up as I continue to clean up the relevant code.


stevebarrau (approved):
I understand this PR freezes the current behavior in tests, and we will address the `TODO` issue in subsequent work. Thanks a lot for this test code @illicitonion

PR #52, opened 2022-07-15
Update the rules_jvm_external rules tp latest release 

tjoneslo:
This gets us access to the packages in the maven_install.json file, meaning we no longer need the maven_manifest.json as a seperate file/generation step. This is the first stage of this.


illicitonion:
I think the `maven_manifest.json` file in here should be deleted? Otherwise LGTM!

illicitonion (approved):
Thanks!

PR #54, opened 2022-07-21
Fix up helper library generation

illicitonion:
Where libraries are generated, they should have sources. Where they have
no sources, they shouldn't be generated. And where we're generating a
suite, which itself generates a helper library, we shouldn't generate
one at all.


stevebarrau (approved):
Thanks for this cleanup.

PR #55, opened 2022-07-22
Deduplicate and sort imports

illicitonion:
This avoids generating a bunch of duplicated imports. It also fixes up
the fact that we were previously leaking packages across java_test
targets in module mode (as evidenced by the removals of the
non-TODO-annotated lines in
java/gazelle/testdata/java_test_per_file/src/test/java/com/example/modulewithhelpers/BUILD.want).


java/gazelle/private/sorted_set/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "sorted_set",
+    srcs = ["btreeset.go"],
+    importpath = "github.com/bazel-contrib/rules_jvm/java/gazelle/private/sorted_set",
+    visibility = ["//visibility:public"],
```
stevebarrau, line 7:
```suggestion
    visibility = ["//java/gazelle:__subpackages__"],
```

java/gazelle/private/javaparser/BUILD.bazel
```diff
@@ -11,6 +11,7 @@ go_library(
         "//java/gazelle/private/java",
         "//java/gazelle/private/javaparser/netutil",
         "//java/gazelle/private/javaparser/proto/gazelle/java/javaparser/v0:javaparser",
+        "//java/gazelle/private/sorted_set",
```
stevebarrau, line 14: Can we also correct the visibility here to `//java/gazelle:__subpackages__`? We choose to name it private after other bazel project instead of internal and so gazelle_go does not help.

stevebarrau (approved):
Thanks a lot for those corrections.

PR #58, opened 2022-07-28
Reduce generate to a single codepath

illicitonion:
Before, code was duplicated across the module/non-module and
test-per-file/test-per-suite axes, which led to inconsistencies and
made it hard to modify the behaviour across the axes. It was also hard
to reason about what the essential differences were across the axes,
because completely independent code was responsible for generation in
each.

Instead, have a single codepath, with conditional logic to conditionally
accumulate the differences in common containers, and express what to
generate as a result of those common containers.

Also:
* Make more clear to what root assorted paths are relative.
* Use parameterised factory functions for each rule generation.
* Set testonly on all test helper libraries, not just in non-module
  mode.
* Don't set broad visibility for test helper libraries.
* Properly set testonly packages on non-module mode test helper
  libraries.
* Generate helper libraries, and helper-only test suites, if helpers but
  no tests are present.

Note that in module mode, java_binary targets are now generated before
java_library targets, which is consistent with the behaviour in
non-module mode. This accounts for most of the test churn.

### DIFF
--- a/java/gazelle/generate.go
+++ b/java/gazelle/generate.go
@@ -10,7 +10,6 @@ import (
 	"unicode"
 
 	"github.com/bazel-contrib/rules_jvm/java/gazelle/javaconfig"
-	"github.com/bazel-contrib/rules_jvm/java/gazelle/private/java"
 	"github.com/bazel-contrib/rules_jvm/java/gazelle/private/javaparser"
 	"github.com/bazel-contrib/rules_jvm/java/gazelle/private/maven"
 	"github.com/bazel-contrib/rules_jvm/java/gazelle/private/sorted_set"
@@ -25,15 +24,13 @@ type main struct {
 }
 
 type javaFile struct {
-	path string
-	pkg  string
+	pathRelativeToBazelWorkspaceRoot string
+	pkg                              string
 }
 
-type javaFiles []javaFile
-
-func (x javaFiles) Len() int           { return len(x) }
-func (x javaFiles) Less(i, j int) bool { return x[i].path < x[j].path }
-func (x javaFiles) Swap(i, j int)      { x[i], x[j] = x[j], x[i] }
+func javaFileLess(l, r javaFile) bool {
+	return l.pathRelativeToBazelWorkspaceRoot < r.pathRelativeToBazelWorkspaceRoot
+}
 
 // GenerateRules extracts build metadata from source files in a directory.
 //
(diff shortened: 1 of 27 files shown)


java/gazelle/generate.go
```diff
@@ -25,15 +24,13 @@ type main struct {
 }
 
 type javaFile struct {
-	path string
-	pkg  string
+	pathRelativeToBazelWorkspaceRoot string
```
stevebarrau, line 27:
```suggestion
	path string
```

java/gazelle/generate.go
```diff
@@ -100,146 +97,125 @@ func (l javaLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 		res.Imports = append(res.Imports, []string{})
 	}
 
-	javaFilenames := filterStrSlice(args.RegularFiles, func(f string) bool { return filepath.Ext(f) == ".java" })
-	if len(javaFilenames) == 0 {
-		if isModule && cfg.IsModuleRoot() {
-			l.generateModuleRoot(args, cfg, &res)
-		}
+	javaFilenamesRelativeToPackage := filterStrSlice(args.RegularFiles, func(f string) bool { return filepath.Ext(f) == ".java" })
```
stevebarrau, line 100:
```suggestion
	javaFilenames := filterStrSlice(args.RegularFiles, func(f string) bool { return filepath.Ext(f) == ".java" })
```

java/gazelle/generate.go
```diff
@@ -254,166 +230,128 @@ func (l javaLang) generateModuleRoot(args language.GenerateArgs, cfg *javaconfig
...
-					res,
-				)
-			}
-		} else {
-			if hasTest {
-				var srcs []string
-				for _, tf := range javaFilenames {
-					srcs = append(srcs, strings.TrimPrefix(tf.path, args.Rel+"/"))
-				}
+func generateJavaLibrary(pathToPackageRelativeToBazelWorkspace string, name string, srcsRelativeToBazelWorkspace []string, packages []string, imports []string, testonly bool, res *language.GenerateResult) {
```
stevebarrau, line 308:
```suggestion
func generateJavaLibrary(path string, name string, srcs, packages, imports []string, testonly bool, res *language.GenerateResult) {
```

java/gazelle/generate.go
```diff
@@ -254,166 +230,128 @@ func (l javaLang) generateModuleRoot(args language.GenerateArgs, cfg *javaconfig
...
+	} else {
+		r.SetAttr("visibility", []string{"//:__subpackages__"})
 	}
+	r.SetPrivateAttr(packagesKey, packages)
+	res.Gen = append(res.Gen, r)
+	res.Imports = append(res.Imports, imports)
+}
+
+func generateJavaBinary(m main, libName string, res *language.GenerateResult) {
+	r := rule.NewRule("java_binary", m.className) // FIXME check collision on name
```
stevebarrau, line 329:
```suggestion
	r := rule.NewRule("java_binary", m.className) // FIXME check collision on name in module mode
```

java/gazelle/generate.go
```diff
@@ -254,166 +230,128 @@ func (l javaLang) generateModuleRoot(args language.GenerateArgs, cfg *javaconfig
...
+	r.SetAttr("main_class", m.pkg+"."+m.className)
+	r.SetAttr("runtime_deps", []string{":" + libName})
+	r.SetAttr("visibility", []string{"//visibility:public"})
+	res.Gen = append(res.Gen, r)
+	res.Imports = append(res.Imports, []string{})
 }
 
-func makeSingleJavaTest(f javaFile, testutilPackageNames *sorted_set.SortedSet[string], imports *sorted_set.SortedSet[string], res *language.GenerateResult) {
-	testName := strings.TrimSuffix(f.path, ".java")
+func generateJavaTest(pathToPackageRelativeToBazelWorkspace string, f javaFile, includePackageInName bool, imports *sorted_set.SortedSet[string], res *language.GenerateResult) {
```
stevebarrau, line 337:
```suggestion
func generateJavaTest(path string, f javaFile, includePackageInName bool, imports *sorted_set.SortedSet[string], res *language.GenerateResult) {
```

stevebarrau (approved):
Thanks for this much cleaner refactor Daniel.

Can you shorten some of the identifiers in the PR though.

illicitonion:
Discussed identifier length offline - going to leave them long for now, and possibly revisit how we make relative-path root differences clear in the future.

PR #61, opened 2022-08-03
java_test_suite: Deduplicate dependencies

illicitonion:
Previously, if someone had manually added a dependency on the helper
library, they would get an error because of a duplicate dependency.

Also, they may end up with a dependency loop of the helper library on
itself.

Avoid both of these issues.

### DIFF
--- a/java/private/create_jvm_test_suite.bzl
+++ b/java/private/create_jvm_test_suite.bzl
@@ -72,19 +72,24 @@ def create_jvm_test_suite(
     nontest_srcs = [src for src in srcs if not _is_test(src, test_suffixes)]
 
     if nontest_srcs:
+        lib_dep_name = "%s-test-lib" % name
+        lib_dep_label = ":%s" % lib_dep_name
+        deps_for_library = [dep for dep in deps or [] if _absolutify(dep) != _absolutify(lib_dep_label)]
+
         # Build a shared test library to use for everything. If we don't do this,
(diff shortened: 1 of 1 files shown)


java/private/create_jvm_test_suite.bzl
```diff
@@ -72,19 +72,23 @@ def create_jvm_test_suite(
     nontest_srcs = [src for src in srcs if not _is_test(src, test_suffixes)]
 
     if nontest_srcs:
+        lib_dep_label = ":%s-test-lib" % name
```
tjoneslo, line 75: could you create a `lib_dep_name` and use it in the `define_library()` as well. Fewer duplicate of the same code.

java/private/create_jvm_test_suite.bzl
```diff
@@ -113,3 +117,34 @@ def create_jvm_test_suite(
...
+
+def _absolutify(label_str):
+    repo = ""
+    package = ""
+    name = ""
+    if not label_str.startswith("@"):
+        repo = ""
+    else:
+        parts = label_str.split("//")
+        repo = parts[0]
```
tjoneslo, line 137: this looks backwards and hard to read
```suggestion
     if label_str.startswith("@"):
        parts = label_str.split("//")
         repo = parts[0]
```
and rely on the initial setting of repo to blank to carry through.

tjoneslo (approved):
two small comments about code style otherwise LGTM

PR #63, opened 2022-08-04
Delete non-server mode

illicitonion:
In reality we only use server mode.

Non-server mode can be emulated by driving with gRPC calls if needed,
but the logic itself isn't valuable, but adds maintenance burden and
confusion.


tjoneslo (approved):
Less code is always better

PR #66, opened 2022-08-08
Report spotbugs issues to stderr not stdout

### LINKED ISSUE #65: Should lint violations be printed to stderr?
We use the bazel event stream in CI to identify why a bazel build failed. It primarily relies on stderr containing the error, but for `spotbugs_test`, the output is printed to stdout.

https://github.com/bazel-contrib/rules_jvm/blob/da93854cba5575e87a3efd76202ce39bf7eb70e9/java/private/spotbugs.bzl#L42

I was wondering if you'd consider using stderr instead?

illicitonion:
Fixes #65


shs96c (approved):
LGTM

PR #67, opened 2022-08-10
Use java tools parser

tjoneslo:
This removes the JavaParser library as the Java parser for the Gazelle java plugin. it also removes several other pieces of dead code from the original command line parser origination of this project. This strips down the Java server process a great deal. 

This now uses the javac parser that comes with the JDK. This avoids the problems we encountered with the parsing the newer Java (17 or later) language features. 

The output from the parser no longer includes the "java.lang.*" imports and any of the other implied import classes. It also no longer attempts to parse through the entire project looking for classes to resolve the imports. Nor does it try to parse the jars looking for the imports to resolve the classes. Since this work is done by the Gazelle side, we let Gazelle do this work.


illicitonion (approved):
LGTM, thanks!

PR #70, opened 2022-08-26
Reworking Java server lifecycle management

illicitonion:
Reworking Java server lifecycle management

* Rather than having a watcher sniffing for gRPC traffic, and shutting
  down the server when it doesn't see any, have the server shut itself
  down if it's been idle for a while. This has a few advantages:
   * It means that if a single package takes longer than the deadline
     to parse, we won't abruptly kill the server - we now stay alive
     when there are in-flight requests.
   * It also means that if we end up with lingering server processes
     (e.g. beacuse of an error on the Go side), they will clean
     themselves up at some point.
   * It means the server will exit cleanly, rather than seeing a signal.
* Remove the javaparser-wrapper completely. Right now, this extra
  process adds complexity to what's going on (multiple processes,
  multiple ports), as well as a small amount of overhead proxying all
  requests.
* Move same-package import filtering from the javaparser-wrapper into
  generate.go directly.
* Shut down when Gazelle tells us we're done parsing source.
  This hook was only recently added to Gazelle, so if people are using
  and older version of gazelle, the server will time itself out.
* Have the Java process pick its own port and report it back, rather
  than be given a port which is _hopefully_ free - this avoids a TOCTOU
  race.

This has a few benefits:
1. It simplifies the general flow.
2. It seems to speed up running gazelle in this repo on my machine from
   generally taking ~3.8 seconds to generally taking ~1.8 seconds.


java/gazelle/private/servermgr/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+
+go_library(
+    name = "servermgr",
```
shs96c, line 4: Nit: `mgr` is a lazy name. In the grand scheme of things, it doesn't matter, but `manager` is preferable IMHO :)

Feel free to do what you will with this feedback. It's not a blocker at all.

java/gazelle/private/servermgr/servermgr.go
```diff
@@ -0,0 +1,130 @@
...
+	if err := cmd.Start(); err != nil {
+		return nil, fmt.Errorf("failed to start javaparser sever: %w", err)
+	}
+
+	port, err := readPort(portFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read port from javaparser server - maybe it crashed: %w", err)
+	}
+
+	addr := fmt.Sprintf("127.0.0.1:%d", port)
```
shs96c, line 77: I wonder what happens on an IPv6-only system? `localhost`, maybe?

shs96c (approved):

java/gazelle/private/servermgr/BUILD.bazel
illicitonion: So goes the go way, alas. But I completely agree, and I will take this opportunity to fight back :D Done

java/gazelle/private/servermgr/servermgr.go
illicitonion: Done!

PR #75, opened 2022-08-29
Migrate tests to use Gazelle's integration test runner

illicitonion:
Advantages:
* allows us to delete a lot of fragile code re-implementing Gazelle's
  internal workfows.
* Each top level testdata directory is a standalone test target which
  can run in parallel, and is independently cached.
* Stdout, stderr, and exit code are assertable artifacts of these tests.
  This is particularly good because right now we have stderr we probably
  don't expect, and this makes that hard to miss. (We can fix up
  removing that stderr in the future).
* Performs a full resolve, rather than a partial one (which means we
  actually see the generated BUILD files as a user would). Several of
  the changes in this PR are adding dependencies as a result of this.
* No longer uses a fake resolver. We can fake things out with
  `# gazelle:resolve directives` or with an actual maven_install.json if
  we wish.
* Allows easily passing flags to gazelle.

Drawbacks:
* We can no longer assert on the output of generate, only resolve. I
  think this is probably fine - if we want to assert on detected
  imports, that should be probably be done at a Java unit test level
  anyway.
* Currently each of the test targets takes 30 seconds to complete. This
  is because we wait for the Java server to terminate before the test is
  considered complete, and we don't proactively kill it.
  https://github.com/bazel-contrib/rules_jvm/pull/70 and
  https://github.com/bazelbuild/bazel-gazelle/pull/1325 will address
  this.

Open questions:
* Which tests do we want to have a maven_install.json for vs use
  `# gazelle:resolve` for vs just assert there's a stderr log?
* What's the future of `resolve_test`? I don't think it's super valuable
  as-is, but for now I've just preserved it (and moved some helpers into
  it) - let's have this discussion post-merge.

How to review this change:
* All of the BUILD.want.gen files are deleted.
* BUILD.want.res has been renamed BUILD.out - any content changes here
  are because our test pipeline is more faithful to reality than the old
  one was.
* Look at each of the `expectedStderr.txt` files and if anything looks
  terribly offensive, we should fix it.


java/gazelle/private/testutil/reset_zerolog_timestamps/dummy_language.go
```diff
@@ -0,0 +1,13 @@
+package reset_zerolog_timestamps
```
blorente, line 1: Nit: Could you rename this file to "placeholder_language" or "noop_language" (for inclusive language)

java/gazelle/testdata/annotations/expectedStderr.txt
```diff
@@ -0,0 +1 @@
+[90m12:00AM[0m [31mWRN[0m [1mjava/gazelle/private/maven/resolver.go:33[0m[36m >[0m not loading maven dependencies [36merror=[0m[31m"open %WORKSPACEPATH%/maven_install.json: no such file or directory"[0m [36m_c=[0mmaven-resolver
```
blorente, line 1: It would be nice to have a record of how this was generated, so that it can be updated in the future.

java/gazelle/testdata/BUILD.bazel
```diff
@@ -0,0 +1,29 @@
...
+[
+    gazelle_generation_test(
+        name = file[0:-len("/WORKSPACE")],
+        gazelle_binary = ":gazelle",
+        gazelle_timeout_seconds = 30,
+        test_data = glob(
+            include = [file[0:-len("/WORKSPACE")] + "/**"],
+        ),
+    )
+    for file in glob(["**/WORKSPACE"])
```
blorente, line 28: I don't usually like iterators on BUILD files, but I quite like this one :D

java/gazelle/testdata/BUILD.bazel
```diff
@@ -0,0 +1,29 @@
...
+    visibility = ["//visibility:private"],
+)
+
+[
+    gazelle_generation_test(
+        name = file[0:-len("/WORKSPACE")],
+        gazelle_binary = ":gazelle",
+        gazelle_timeout_seconds = 30,
+        test_data = glob(
+            include = [file[0:-len("/WORKSPACE")] + "/**"],
```
blorente, line 25: If I understand correctly, this will not glob sub packages, which may lead to some tricky-to-debug situations.

If we want to keep the invariant that "No testdata repo should contain BUILD files", we should have a linter or write it down in the readme.

repositories.bzl
```diff
@@ -59,10 +59,14 @@ def contrib_rules_jvm_gazelle_deps():
...
+        sha256 = "b794b5c4af78574e4bf574c6b2336bbffbdfeadfaae1c14c6184f154e44fdb47",
+        strip_prefix = "bazel-gazelle-622d8889c63227a71d6b393ba5e3a8b8a6761466",
         urls = [
-            "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.26.0/bazel-gazelle-v0.26.0.tar.gz",
-            "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.26.0/bazel-gazelle-v0.26.0.tar.gz",
+            "https://github.com/bazelbuild/bazel-gazelle/archive/622d8889c63227a71d6b393ba5e3a8b8a6761466.tar.gz",
+        ],
+        patch_args = ["-p1"],
+        patches = [
+            "@contrib_rules_jvm//java/gazelle/private/patches:bazel_gazelle-1324-allow-configuring-timeout-of-generation-tests.patch",
```
blorente, line 69: Should we upstream this? If we have already, could you include a reference to the PR/issue in a comment?

blorente (approved):
Love to see this code go away.
Added some comments and I think it looks good, but the change is large enough that I think it warrants a second opinion.

blorente (approved):

PR #78, opened 2022-08-30
Allow setting test attributes based on annotations

illicitonion:
This allows for registering that e.g.

```java
@FlakyTest
class SomeTest {}
```

should have `flaky = True` generated.

In suite mode, separate targets will be generated for these special
targets.


java/gazelle/testdata/attribute_setting/arguments.txt
```diff
@@ -0,0 +1 @@
+-java-annotation-to-attribute=com.example.annotation.FlakyTest=flaky=True
```
stevebarrau, line 1: Is this something that would come from gazelle directive? Can we thus set this in the build file instead?

stevebarrau (commented):
This is a neat idea. I did not think we would get to this level of smart in this extension this soon. Is this something we could add documentation for in https://github.com/bazel-contrib/rules_jvm/tree/main/java/gazelle#configuration-options; hopefully this helps discover this new feature.

java/gazelle/testdata/attribute_setting/arguments.txt
illicitonion: Sadly not currently because of https://github.com/bazelbuild/bazel-gazelle/issues/1313

I'd love to move this to be a directive-driven piece of config, rather than flag-driven, but there's still a decent amount of upstream discussion to have there.

illicitonion:
> Is this something we could add documentation for in https://github.com/bazel-contrib/rules_jvm/tree/main/java/gazelle#configuration-options; hopefully this helps discover this new feature.

Yep - added a general reference to look at that file for config. Ideally I'd like not to duplicate the docs in the two files, but we can if you'd prefer.

stevebarrau (approved):

PR #79, opened 2022-09-01
Allow setting wrapper targets based on annotations

illicitonion:
This allows for registering that, e.g.

```java
@RequiresNetwork
class SomeTest {}
```

should be generated as:

```starlark
requires_network(
    java_test,
    ...
)
```

instead of:

```starlark
java_test(
    ...
)
```

In suite mode, separate targets will be generated for these special
targets.


illicitonion:
Closing this for now - will revive if it becomes necessary

stevebarrau (approved):

PR #81, opened 2022-09-01
Fully qualified static method calls

tjoneslo:
Add functionality to find and process a fully qualified static method call. This may be a reference to a package/class which does not exist anywhere else, so make sure to add this to the potential list of classes we've found for processing.


illicitonion (changes requested):
The extra warnings in the stderr of the testdata are kind of concerning - they point at classes being detected which are definitely not things that should be imported. Could you tweak the heuristics to avoid the false positives? Maybe verifying that the first token of `package` is a lower-case identifier character?

illicitonion (approved):
LGTM, thanks! If you can fmt (looks like a java comment needs wrapping?) I'll gladly merge!

PR #83, opened 2022-09-01
Add classifier support to maven targets

illicitonion:


shs96c (approved):
LGTM. The equivalent code for what you're doing in `rules_jvm_external` is found here: https://github.com/bazelbuild/rules_jvm_external/blob/master/private/rules/maven_utils.bzl#L1-L25 Here they call `classifier` `type`, but the idea is the same.

I don't think the download will expand to cope with more than what you have here, so there's nothing else to be done.

PR #85, opened 2022-09-15
Also detect JUnit Pioneer imports imply JUnit 5

rdesgroppes:
In the same way that #80 brought detection of test classes that only use `@ParameterizedTest` to determine they're also JUnit 5-dependent, the change proposed here aims at extending it to test annotations provided by [JUnit Pioneer](https://junit-pioneer.org/) (Eclipse Public License v2.0).

As its classes are organized under various [sub-packages](https://javadoc.io/doc/org.junit-pioneer/junit-pioneer/latest/index.html), only the prefix of the imports is considered here.

Suggestion: perhaps we could consider a good old regex to combine all these predicates into one.

### DIFF
--- a/java/gazelle/generate.go
+++ b/java/gazelle/generate.go
@@ -453,11 +453,19 @@ func importsJunit4(imports *sorted_set.SortedSet[string]) bool {
 	return imports.Contains("org.junit.Test") || imports.Contains("org.junit")
 }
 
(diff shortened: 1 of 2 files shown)


illicitonion (approved):
Thanks!

rdesgroppes:
You're welcome!

PR #89, opened 2022-09-28
Avoid Java 17 printing a warning about the SecurityManager

shs96c:
One of the planned clean-ups of Java that's happening is the removal of the SecurityManager. The plan is to replace it with utilities and methods that provide similar capabilities but in a more controlled way.

In order to make this process smoother, the JRE now prints a warning when `System.setSecurityManager()` is called.

A common use-case for the Security Manager is to prevent people calling `System.exit()` in untrusted code. Indeed, this particular case is called out in the JEP:
https://openjdk.org/jeps/411

Now, people using the JUnit5 runner don't really deserve to see this warning. There's nothing they can do about it, and we're fully aware of the impending System.exit apocalypse. Worse, many of users are still on Java 11, and setting the security manager there is totally legit behaviour, and in no way frowned upon.

Fortunately, Java 17 offers a system property we can call to disable the warning. That system property is `java.security.manager` with a value of `allow`. Java 17 knows that `allow` is a fine thing to say, and simply disables the warning about the security manager.

The bad news is that in Java 11 the _very self-same system property_ is used to set the security manager. In fact (!) it's used for that in Java 17 too, but only if the value isn't one of `default`, `disallow`, or `allow`. The way it does this is by passing in a class name.

So! In Java 11, calling `-Djava.security.manager=allow` tells the JRE to set the security manager to the class who's name is `allow`. You can guess how well that goes down.

Well then. That's fine. We'll just use the name of our own security manager as the value of the system property, and we're all good, right? Except our security manager was never really designed to be used this way. Much of this PR is just shuffling code around so that we can use our own security manager as the default security manager.

So, how's your yak?

### DIFF
--- a/java/private/junit5.bzl
+++ b/java/private/junit5.bzl
@@ -47,12 +47,22 @@ def java_junit5_test(name, test_class = None, runtime_deps = [], **kwargs):
     else:
         clazz = get_package_name() + name
 
+    jvm_flags = kwargs.pop("jvm_flags", [])
+    for f in jvm_flags:
+        if f.startswith("-Djava.security.manager="):
+            fail("Only the JUnit5 runner is allowed to set the security manager via a JVM flag")
+
     java_test(
         name = name,
         main_class = "com.github.bazel_contrib.contrib_rules_jvm.junit5.JUnit5Runner",
         test_class = clazz,
         runtime_deps = runtime_deps + [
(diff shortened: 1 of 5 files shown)


shs96c:
This PR depends on #88 in order to finish wiring up the security manager properly.

illicitonion (approved):

rdesgroppes:
Doesn't work for me, but a quick tweak on top does it. Kindly have a look at #92.

PR #90, opened 2022-09-29
Suppress warnings about security manager on java 17

shs96c:
Have another crack at sorting out the warning printed by the JRE when using Java 17+ and running a junit5 test.

The previous fix prevented a warning being displayed at runtime, but then caused the JRE to emit a warning that the `SecurityManager` had been set on the command line.

To avoid that, we now avoid setting the `SecurityManager` on the command line, and we're back to setting it in code.

To avoid a warning being displayed, we now "simply" check the version of Java we're using at runtime. If it's Java 11, then we use the `SecurityManager` as we did in the first place. On Java 17, however, we make a series of poorly considered choices that tie us to the OpenJDK to reach into the internals of `System` and tell it that we've already registered a `SecurityManager` instance. This prevents the warning message from being displayed.

On the off-chance that we're running on a JRE that has already had `SecurityManager` removed (I hear that may be as soon as Java 20), we provide a fall-back in the shape of a no-op control that does nothing.

All this to prevent a test calling `System.exit`....

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/JUnit5Runner.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/JUnit5Runner.java
@@ -56,7 +56,9 @@ public class JUnit5Runner {
     }
 
     // Load the java 17 toggle by reflection, because it's tied
-    // so closely to the OpenJDK
+    // so closely to the OpenJDK (it relies on the internal fields
+    // of both `sun.misc.Unsafe` and `java.lang.System`: it's a
+    // gross hack.
     try {
       var java17ToggleClazz =
           Class.forName(JAVA17_SYSTEM_EXIT_TOGGLE).asSubclass(SystemExitToggle.class);
@@ -66,6 +68,7 @@ public class JUnit5Runner {
       // this would be a combination of `ReflectiveOperationException` and
       // `SecurityException`, but the latter is scheduled for removal so
       // relying on it seems like a Bad Idea.
+      System.err.println("Failed to load Java 17 system exit override: " + e.getMessage());
 
       // Fall through
     }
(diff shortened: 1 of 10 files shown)


java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/JUnit5Runner.java
```diff
@@ -13,32 +13,68 @@ public class JUnit5Runner {
...
+
+  private static SystemExitToggle getSystemExitToggle() {
+    var version = Runtime.version();
+
+    if (version.feature() < 12) {
+      return new Java11SystemExitToggle();
+    }
+
+    // Load the java 17 toggle by reflection, because it's tied
+    // so closely to the OpenJDK
```
illicitonion, line 59:
```suggestion
    // Load the java 17 toggle by reflection, because it's tied
    // so closely to the OpenJDK.
    // If we import it, it will force `sun.misc.Unsafe` to be class-loaded, which wouldn't work on non-OpenJDKs.
```

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/JUnit5Runner.java
```diff
@@ -13,32 +13,68 @@ public class JUnit5Runner {
...
+    try {
+      var java17ToggleClazz =
+          Class.forName(JAVA17_SYSTEM_EXIT_TOGGLE).asSubclass(SystemExitToggle.class);
+      return java17ToggleClazz.getDeclaredConstructor().newInstance();
+    } catch (Exception e) {
+      // We don't care _why_ we can't load the toggle, but we can't. Ideally
```
illicitonion, line 65: Let's at least print out `e.getMesage()`?

illicitonion (approved):
LGTM, two small comments :)

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/JUnit5Runner.java
shs96c: We also crack open `System` in a way that relies on the OpenJDK implementation of that class. I'll add more text to the comment.

PR #91, opened 2022-09-30
Add `junit-platform-suite` to runner filters

rdesgroppes:
This is to allow [@Suite](https://junit.org/junit5/docs/current/api/org.junit.platform.suite.api/org/junit/platform/suite/api/Suite.html) annotations to be honored when users need to leverage the [JUnit Platform Suite Engine](https://junit.org/junit5/docs/current/user-guide/#junit-platform-suite-engine) (as long as they pass the necessary runtime dependencies as documented there, of course).

Adding `junit-platform-suite` to the runner filters allows in particular that classes listed by `@SelectClasses` get actually taken into account (among other features of the JUnit Platform Suite Engine).

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/ActualRunner.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/ActualRunner.java
@@ -43,7 +43,7 @@ public class ActualRunner implements RunsTest {
       var request =
           LauncherDiscoveryRequestBuilder.request()
               .selectors(List.of(classSelector))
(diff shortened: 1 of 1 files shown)


shs96c (approved):
Thank you for the PR. The tests are green, and this seems sensible. Merging! This is much appreciated!

rdesgroppes:
@shs96c What's even more appreciated is your prompt (and positive) review, thanks!

marcphilipp:
@shs96c Why is there a default engine filter in place? Could this instead be solved via configuration on `java_test_suite` and `java_junit5_test`?

marcphilipp:
Follow-up: #106

PR #93, opened 2022-10-01
Fix test filter greediness for method names

rdesgroppes:
It's a bit annoying to designate a test target and have neighboring targets hit as well, especially while debugging because breakpoints are not reached under the desired conditions.

The present change aims at preventing the `MyTest#testSomething` pattern from matching the `testSomethingElse` method in the following example:
```java
public class MyTest {
  @Test
  void testSomething() {
  }

  @Test
  void testSomethingElse() {
  }
```

With `asPredicate`, the test filter matches test methods _starting with_ the given method name:
> Creates a predicate that tests if this pattern is found in a given input string.

Unless the pattern doesn't encompass a method name, appending a `$` to the input pattern makes sure only exact method names are considered.

Note: think about filter greediness for class names.

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/PatternFilter.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/PatternFilter.java
@@ -2,6 +2,8 @@ package com.github.bazel_contrib.contrib_rules_jvm.junit5;
 
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.junit.platform.engine.FilterResult;
 import org.junit.platform.engine.TestDescriptor;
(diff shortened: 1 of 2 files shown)


rdesgroppes:
@shs96c really no hurry for this one: better safe than sorry...

shs96c (approved):
I've been thinking, and I'm going to land this PR. Being able to actually run what you expect is a nice feature. Maybe sticking too closely to what the existing test runner does isn't going to be the most helpful thing for users.

I think we'll need to document how to use test filtering properly, but I don't think that's a reason to hold back this PR.

Thank you for this!

shs96c:
But could you please rebase?

rdesgroppes:
Will do within max 3 days.

shs96c:
That's very much appreciated. Thank you!

rdesgroppes:
Done.

shs96c (approved):
LGTM! Thank you for the PR!

PR #94, opened 2022-10-02
Add dynamic test results to XML output

rdesgroppes:
This is to add results of `@Parameterized` and other "dynamic" tests to the XML output file, otherwise these don't show up in IDE views (eg. IntelliJ IDEA `Bazel test` runs) nor in CI reports.

In the following example:
```java
import org.junit.jupiter.api.Test;
import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.ValueSource;

public class SomeTest {
    @Test
    void shouldAlwaysPass() {
    }

    @ParameterizedTest
    @ValueSource(ints = {0, 2, 4})
    void shouldBeEven(int value) {
    }
}
```
... we would get:
| prior to this change | with this change |
| -- | -- |
| image | image |

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BazelJUnitOutputListener.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BazelJUnitOutputListener.java
@@ -71,6 +71,11 @@ public class BazelJUnitOutputListener implements TestExecutionListener, Closeabl
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


illicitonion (approved):
Looks great, thanks!

Ideally I'd love a test for this behaviour, but I don't think we currently have any existing tests for the XML output format, and adding the first one seems like an unnecessary burden on you :)

PR #98, opened 2022-10-07
Allow finding and processing a classname starting with an underscore

tjoneslo:
This is an unusual corner case, but we've found that some class names can begin with an underscore. Make sure that the Java import processor can correctly identify these, separate them properly, and update the dependencies as needed.

### DIFF
--- a/java/gazelle/private/java/import.go
+++ b/java/gazelle/private/java/import.go
@@ -19,7 +19,7 @@ func NewImport(imp string) *Import {
 		if unicode.IsUpper(rune(parts[i][0])) {
 			break
 		}
-        // Assume that a _ prefix denotes a class not a package.
+		// Assume that a _ prefix denotes a class not a package.
 		// This isn't a strong convention in the Java community, but has been spotted in the wild.
 		// If it breaks (i.e. we discover packages starting with underscores), we may tweak this heuristic further.
 		//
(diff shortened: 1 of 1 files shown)


shs96c (approved):

shs96c:
OSRB ok to upstream

java/gazelle/private/java/import.go
```diff
@@ -19,6 +19,12 @@ func NewImport(imp string) *Import {
 		if unicode.IsUpper(rune(parts[i][0])) {
 			break
 		}
+		// The "_TESTONLY" comes from the parser to identify cases where there is both a
+		// java_library and a java_test_suite in the same file, and a package may need access
+		// to the test library. see java/gazelle/generate.go for details.
+		if strings.HasPrefix(parts[i], "_") && !strings.HasPrefix(parts[i], "_TESTONLY") {
+		    break
+		}
```
illicitonion, line 27:
```suggestion
		// Assume that a _ prefix denotes a class not a package.
		// This isn't a strong convention in the Java community, but has been spotted in the wild.
		// If it breaks (i.e. we discover packages starting with underscores), we may tweak this heuristic further.
		//		
		// The "_TESTONLY" comes from the parser to identify cases where there is both a
		// java_library and a java_test_suite in the same file, and a package may need access
		// to the test library. see java/gazelle/generate.go for details.
		if strings.HasPrefix(parts[i], "_") && !strings.HasPrefix(parts[i], "_TESTONLY") {
		    break
		}
```

WDYT about this addition?

illicitonion (approved):
LGTM, but one suggestion

PR #99, opened 2022-10-11
Support multiple classes and/or methods in filter

rdesgroppes:
This is to allow the selection of:
1. a test class holding nested test class(es), either `static` or explicitly `@Nested`, in which case IntelliJ IDEA's Blaze plugin turns out to pass something like:
   ```
   --test_filter=path.to.SomeTest,path.to.SomeTest.NestedTest
   ```
   ... where the `,` only needs to be replaced by a `¦`.
2. multiple test methods specified through the "Rerun Failed Tests" feature of IntelliJ IDEA, in which case Blaze passes something like:
   ```
   --test_filter=path.to.SomeTest#testFailed1,testFailed2,path.to.SomeTest.NestedTest#testFailed3
   ```
   ... where, in addition to replacing `,`s by `¦`s, `path.to.SomeTest#` (the preceding class name) needs to be propagated to `testFailed2`.

Prior to this change, in both cases, IntelliJ IDEA would report:
> No tests were found

### DIFF
--- a/java/test/com/github/bazel_contrib/contrib_rules_jvm/junit5/FilteringTest.java
+++ b/java/test/com/github/bazel_contrib/contrib_rules_jvm/junit5/FilteringTest.java
@@ -6,6 +6,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import java.lang.reflect.Method;
 import java.util.Optional;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.engine.config.DefaultJupiterConfiguration;
 import org.junit.jupiter.engine.config.JupiterConfiguration;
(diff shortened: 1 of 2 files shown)


java/test/com/github/bazel_contrib/contrib_rules_jvm/junit5/FilteringTest.java
```diff
@@ -85,6 +129,46 @@ public void shouldIncludeATestMethodIfTheFilterIsJustTheClassName() {
...
+    assertFalse(nestedTestResult.included(), "nested class should not be matched");
+  }
+
+  @Test
+  public void shouldIncludeANestedTestMethodIfTheFilterIsJustTheNestedClassName() {
+    PatternFilter filter =
+        new PatternFilter(JUnit5StyleTest.NestedTest.class.getName().replace("$", "\\$") + "#");
+
+    FilterResult testResult = filter.apply(testMethodTestDescriptor);
+    assertFalse(testResult.included(), "parent class should not be matched");
```
rdesgroppes, line 146:
```suggestion
    assertFalse(testResult.included(), "enclosing class should not be matched");
```

rdesgroppes: Fixed.

shs96c (approved):
Looks good to me. Thank you for the PR!

PR #100, opened 2022-10-12
Avoid asking users to add things to their `maven` workspace

shs96c:
The google code formatter is just for us. Use the correct import to avoid polluting the regular `@maven` workspace.


blorente (approved):
Thanks!

PR #102, opened 2022-10-21
Ensure that the release workflow generates correct release notes

shs96c:
While I'm not convinced we're calculating the SHA256 correctly, the URL and the `strip_prefix` are now better.


blorente (approved):
Same sentiment on the shas, but this is definitely an improvement :) Thanks!

illicitonion (approved):

PR #106, opened 2022-10-29
Allow configuring engine filters

marcphilipp:
This commit introduces `include_engines` and `exclude_engines` arguments
on `java_test_suite` and `java_junit5_test` which can be used to
configure which engines should be used. It replaces the previously
hard-coded list of engines in `ActualRunner`.


marcphilipp:
For context: this allows running test with arbitrary test engines, e.g. jqwik.

shs96c:
I wonder if there's a way to detect the engines that are available without us needing to list them?

shs96c:
Apparently there is, and all we need to do is to use the `ServiceLoader`. I've created #111 to do this.

shs96c:
Do you still want to be able to set these properties? I think we can probably control the engines by changing the dependencies of a test rule....

java/private/create_jvm_test_suite.bzl
```diff
@@ -27,6 +27,8 @@ def create_jvm_test_suite(
         tags = [],
         visibility = None,
         size = None,
+        include_engines = [],
```
shs96c, line 30: We don't need to add these options. They'll be passed through to the underlying `java_junit5_test` via the `**kwargs`

java/private/create_jvm_test_suite.bzl
```diff
@@ -96,6 +100,8 @@ def create_jvm_test_suite(
 
         define_test(
             name = test_name,
+            include_engines = include_engines,
```
shs96c, line 103: These are not needed

java/private/java_test_suite.bzl
```diff
@@ -47,6 +49,8 @@ def java_test_suite(
         deps = None,
         runtime_deps = [],
         size = None,
+        include_engines = [],
```
shs96c, line 52: These can be omitted.

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/JUnit5Runner.java
```diff
@@ -18,6 +20,8 @@ public class JUnit5Runner {
 
   public static void main(String[] args) {
     var testSuite = System.getProperty("bazel.test_suite");
+    List<String> includeEngines = System.getProperty("bazel.include_engines") == null ? null : List.of(System.getProperty("bazel.include_engines").split(","));
```
shs96c, line 23: Please could you move the extraction of the runner-specific values to the `ActualRunner`? That way if we add support for (eg) TestNG or JUnit 6 (should it ever exist) we'll have all the logic localised in one place.

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/JUnit5Runner.java
```diff
@@ -34,7 +38,7 @@ public static void main(String[] args) {
       var constructor =
           Class.forName(JUNIT5_RUNNER_CLASS).asSubclass(RunsTest.class).getConstructor();
       var runsTest = constructor.newInstance();
-      if (!runsTest.run(testSuite)) {
+      if (!runsTest.run(testSuite, includeEngines, excludeEngines)) {
```
shs96c, line 41: Which means that all we need to do is pass in the `testSuite`. These other two parameters can be dropped

shs96c (changes requested):
Again, my apologies that this has taken so long. I genuinely appreciate your patience, and thank you for it.

I've finally had a chance to properly review this, and I'd love to land it now I properly understand what you're doing here. Thank you for your comments about how the engines work, and I've been been doing some extra reading too.

I've left some comments, but these are mostly to do with minimising how the engines feature leaks into other kinds of test. Once those are addressed, I'll happily merge this.

marcphilipp:
@shs96c No worries and thanks for the detailed feedback! I've addressed all your feedback, merged in `main`, and added documentation. If you decide to merge the PR, please squash all commits. PTAL!

shs96c (approved):
Brilliant. Thank you!

PR #108, opened 2022-11-07
Fix up junit reporting

shs96c:
Previously we didn't count the number of disabled or skipped tests and now we do. Adding this highlighted an issue where a disabled test would cause no test output to be created. This was obviously incorrect :)


illicitonion (approved):

stevebarrau:
Thanks for addressing this issue @shs96c. The patch is effective in getting more accurate information about our tests.

In the case of a **disabled test class**, this produces an XML file with no reported test suites. The current Gradle behaviour for something similar is to emits the a test suite and all the individual tests in the disabled test class as skipped. Is this something we could replicate?

PR #109, opened 2022-11-08
[junit5] Also filter dynamically registered tests

rdesgroppes:
This change is a follow-up to [Add dynamic test results to XML output](https://github.com/bazel-contrib/rules_jvm/pull/94), where only partial support for dynamically registered tests was added.

What was missing is that dynamic tests were always matched, disregarding any selected pattern since `TestDescriptor.isTest()` would _always evaluate to `false`_ for them (`isStaticTest` would be more accurate).

Such tests get indeed registered through special "containers":
- either a [TestFactoryTestDescriptor](https://junit.org/junit5/docs/current/api/org.junit.jupiter.engine/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.html),
- or a [TestTemplateTestDescriptor](https://junit.org/junit5/docs/current/api/org.junit.jupiter.engine/org/junit/jupiter/engine/descriptor/TestTemplateTestDescriptor.html).

The fact that both of them (and potentially other `TestDescriptor` implementations) act as _test sources_ could be determined by an additional evaluation of `TestDescriptor.mayRegisterTests()`.

`FilteringTest` methods get here augmented in order to cover a `TestTemplateTestDescriptor` the same way as one of the `TestMethodTestDescriptor`s. Prior to adjusting `PatternFilter.apply`, following test methods were then failing:
- `shouldNotIncludeATestMethodIfTheFilterDoesNotMatchTheMethodName`,
- `ifFilterIsSetButNoTestsMatchTheContainersAreIncluded`,
- `shouldIncludeANestedTestMethodIfTheFilterIsJustTheNestedClassName`,
- `shouldIncludeATestMethodIfTheFilterMatchesTheExactLongMethodName`.

In the below synthetic example, the present change would prevent the `MyTest#testStatic` pattern from matching the `testDynamic` method:
```java
public class MyTest {
  @Test
  void testStatic() {
  }

  @ParameterizedTest
  @ValueSource(booleans = {false, true})
  void testDynamic(boolean flag) {
  }
}
```

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/PatternFilter.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/PatternFilter.java
@@ -34,7 +34,7 @@ public class PatternFilter implements PostDiscoveryFilter {
 
   @Override
   public FilterResult apply(TestDescriptor object) {
-    if (!object.isTest()) {
+    if (!object.isTest() && !object.mayRegisterTests()) {
       return FilterResult.included("Including container: " + object.getDisplayName());
     }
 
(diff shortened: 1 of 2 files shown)


shs96c:
Would it be possible to get a test that verifies this works as expected?

rdesgroppes:
> Would it be possible to get a test that verifies this works as expected?

Sure thing, please hold the line ;)

rdesgroppes:
@shs96c And there you go!

shs96c (approved):
Brilliant. Thank you!

PR #113, opened 2022-11-10
Add a way to exclude maven artifacts from being resolved

### LINKED ISSUE #112: FR: Way to exclude jar from gazelle resolution
Our repository has a list of shared maven dependencies that includes a jar (`"org.openapitools:openapi-generator-cli:6.0.0"`) that gazelle thinks would work for a lot of different packages, and this leads to a lot of `gazelle:resolve` lines in our top-level BUILD files.  It would be nice if there was a `gazelle:do-not-resolve` type of annotation that we could use to exclude that one jar.

chrismgrayftsinc:
Fixes #112

### DIFF
--- a/java/gazelle/configure.go
+++ b/java/gazelle/configure.go
@@ -49,6 +49,7 @@ func (jc *Configurer) KnownDirectives() []string {
 		javaconfig.MavenInstallFile,
 		javaconfig.ModuleGranularityDirective,
 		javaconfig.TestMode,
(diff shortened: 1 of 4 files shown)


chrismgrayftsinc:
Apologies if some of the go is not idiomatic -- I'm a Java programmer primarily these days.

illicitonion (approved):
Perfect, thanks so much!

PR #114, opened 2022-11-10
[junit5] Fix serialization of illegal XML characters

rdesgroppes:
This is to fix invalid XML when a test fails due to an exception whose message contains illegal XML characters, leading to decoding exceptions like:
> org.xml.sax.SAXParseException; lineNumber: 3; columnNumber: 30; An invalid XML character (Unicode: 0x0) was found in the value of attribute "message" and element is "failure".

This may happen when testing invalid protobuf `Message`s, `UUID` bytes, etc.

A simple solution would be to leverage guava's `XmlEscapers` or Apache's `StringEscapeUtils`, but that would infringe:
```
        # The only dependencies here are those required to run
        # a junit5 test. We try not to pollute the classpath, so
        # be very careful when adding new deps here.
```

The present fix therefore mimics what was done in the upstream JUnit 5 project: [Replace illegal chars with their char reference](https://github.com/junit-team/junit5/pull/2275).

Paying some tech debt, the change also adds a test that should have been part of [Fix NPE when translating null exception messages to XML](https://github.com/bazel-contrib/rules_jvm/pull/96) (related).

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestResult.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestResult.java
@@ -97,13 +97,14 @@ class TestResult extends BaseResult {
               return;
             }
 
-            xml.writeAttribute("message", String.valueOf(throwable.getMessage()));
+            xml.writeAttribute(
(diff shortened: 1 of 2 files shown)


illicitonion (approved):
Wonderful, thanks!

PR #116, opened 2022-11-29
Fix time attribute format in JUnit 5 test results

hisener:
This PR sets the root locale for the duration formats in the JUnit test result XMLs since the decimal separator is important for IntelliJ IDEA test reports. Currently, the format depends on the machine’s default locale (e.g., in my case, it was `en_NL`).

The test report with `en_NL` as the default locale before this change:
Screenshot 2022-11-29 at 13 08 53

The test report with `en_NL` as the default locale after this change:
Screenshot 2022-11-29 at 13 09 28

Note that providing a non-parsable value to the [time attribute in `testcase`](https://github.com/bazelbuild/intellij/blob/0edc361d5906f5edbb1b9002cd7ad425a317c89c/base/src/com/google/idea/blaze/base/run/smrunner/BlazeXmlSchema.java#L146) only breaks the duration in test report, assuming its type is string. However, the [time attribute in `testsuite`](https://github.com/bazelbuild/intellij/blob/0edc361d5906f5edbb1b9002cd7ad425a317c89c/base/src/com/google/idea/blaze/base/run/smrunner/BlazeXmlSchema.java#L82) is double, and passing a non-parsable value to it causes `No tests were found` error, even when tests pass successfully.

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestResult.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestResult.java
@@ -4,6 +4,8 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.math.RoundingMode;
 import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
+import java.util.Locale;
 import java.util.Optional;
 import javax.xml.stream.XMLStreamWriter;
 import org.junit.AssumptionViolatedException;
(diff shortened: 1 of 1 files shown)


shs96c:
I think for the sake of repeatability, we want to ensure that the output of the test runner is the same no matter where in the world the tests are being run. I think this is better handled by the IJ plugin.

shs96c:
My apologies. I closed this too hastily. You are, I think, correct to make the fix here.

shs96c:
Thank you for the fix! It's much appreciated, and we'd not noticed the problem.

PR #118, opened 2022-12-31
Add `.co` to list of common package prefixes

robertgates55:
I work for a company called `du.co` and am looking to migrate to bazel. Any chance of adding `.co.` as a common package prefix?


shs96c (approved):
Looks good to me. Thank you for putting this PR together!

Issue #104, opened 2022-10-22
issue with java_test_suite package inference 


m4tta:
So from what I can tell unless your package begins with a common domain `(".com.", ".org.", ".net.", ".io.", ".ai.")` it won't correctly load the class file. 

So for example if I have a test class at `src/test/java/fish/boxjelly/web/WebApplicationTests.java` and has the package of `package fish.boxjelly.web;` It will attempt to load the class using the full path with the `src.test.java.fish.boxjelly.web.WebApplicationTests` instead of `fish.boxjelly.web.WebApplicationTests`

Is there a way to supply a base folder to exclude? If I could tell the test suite that `src.test.java` should be ignored things would work.

PR #120, opened 2023-01-09
Add `.me` to list of common package prefixes

luangong:
I’m working on my personal project and I use `me.luangong.*` as the package prefix. Can we also treat `.me` as a common package prefix? Thanks!

See #118 and #104 for reference.


illicitonion (approved):
Thanks!

PR #121, opened 2023-01-11
fix(#104): Allow custom package prefixes

hjellek:
First of all: Thanks a lot for these rules!

Attempt at making the package prefixes configurable, instead of populating _PREFIXES with all known values.

I need suppport for `.no.` domains for example, and based on #104 there are many more wishes for package prefixes out there.

Feedback and improvements are welcome!


illicitonion (approved):
Thanks for the contribution!

PR #125, opened 2023-01-25
Update the junit-5 dependencies for the test running to the latest vesions

tjoneslo:
This is an attempt to see if the 1.9 version has include/exclude by tags, a feature used by the test suite builder to see if we can get better resolution on the tests.


tjoneslo:
Not currently needed.

PR #127, opened 2023-01-30
gazelle: Add support for rules_jvm_external v2 lockfile

stevebarrau:
Add support for the v2 lockfile format introduced in https://github.com/bazelbuild/rules_jvm_external/pull/822

CC @shs96c


illicitonion (approved):
Thanks!

shs96c (approved):
LGTM. Thanks for making it work with both lock file versions

PR #131, opened 2023-02-01
Revert "Fully qualified static method calls (#81)"

illicitonion:
This reverts commit ea165afe466a1ee3d9d85ac017e4e8b1beb0dfe6.

We've had reports that it breaks untested scenarios.


stevebarrau (approved):
Thanks @illicitonion. I will work with @tjoneslo to provide the tests this broke for us.

shs96c (approved):

PR #132, opened 2023-02-01
Do not fail the build if the security manager is disabled

shs96c:


java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/allow.java
```diff
@@ -0,0 +1,103 @@
...
+ * at runtime. In Java 18, this changed, yet we still want to be able to
+ * block calls to `System.exit` in tests that we execute.
+ *
+ * Fortunately, we can avoid the pain and grief by setting the
+ * `java.security.manager` to `allow` in JDK 12+. Prior to JDK 12+, which
+ * you will note includes the ever-so-popular Java 11, setting this setting
+ * to `allow` will cause the SecurityManager to be set to a class who's
+ * fully-qualified name is `allow`. Hence, this class.
+ *
+ * I can't apologise enough for this being here.
```
blorente, line 16: Once every COBOL update, hacks sprout that end up holding the world together.
The sort that are so deeply integrated into the ducktape of the universe, that if we removed them it would fall apart.
It's a privilege to witness the birth of one such hack.

blorente (approved):
LGTM. This patch oozes jvm black wizardry.

PR #134, opened 2023-02-07
Make the JUnit5 runner compatible with Java 8

fmeum:
This requires getting rid of `var` and polyfilling a few unavailable methods. Since compiling with a fixed `--release` is required to prevent breakages when `SecurityManager` goes away in future releases, this also includes a workaround for https://bugs.openjdk.org/browse/JDK-8214165.


java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BUILD.bazel
```diff
@@ -1,58 +1,46 @@
 load("@rules_java//java:defs.bzl", "java_binary", "java_library")
 load("//java/private:artifact.bzl", "artifact")
 
-# We want to be able to compile the test runner, but we don't want to
-# expose any of its dependencies, since we expect whoever is running
-# tests to add them themselves, and we don't want to pollute the classpath.
-# As such, we do this rather weird thing to make sure we have the runner
-# but none of its dependencies.
-java_import(
```
shs96c, line 9: This was done deliberately so that we don't leak any of our dependencies into people's builds. We want the user to be bringing their version of junit to the table, not accidentally picking up our version

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BUILD.bazel
```diff
@@ -1,58 +1,46 @@
...
-        # someone rebuilds on a JDK that doesn't have it.
-        "--release",
-        "11",
-    ],
-    deps = [
-        # The only dependencies here are those required to run
-        # a junit5 test. We try not to pollute the classpath, so
-        # be very careful when adding new deps here.
+    name = "maven-deps-compile-only",
+    neverlink = True,
```
shs96c, line 42: Won't this still bleed the transitive deps on to the classpath? That's the thing we're trying to avoid.

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/Java17SystemExitToggle.java
```diff
@@ -56,10 +59,15 @@ public void allow() {
...
     // We want to avoid opening java.lang, so let's jump through hoops
 
-    // First problem: we can't call `Unsafe.getUnsafe` directly.
-    var c = Class.forName("sun.misc.Unsafe");
-    var lookup = MethodHandles.privateLookupIn(c, MethodHandles.lookup());
-    var handle = lookup.findStaticVarHandle(c, "theUnsafe", c);
-    return (Unsafe) handle.get();
+    // First problem: we can't call `Unsafe.getUnsafe` directly and it might also be called
+    // differently.
+    Class<?> unsafe = Class.forName("sun.misc.Unsafe");
```
shs96c, line 64: The way this was written before allowed us to avoid needing to open any more java modules. Please revert to the original implementation.

shs96c (changes requested):

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BUILD.bazel
```diff
@@ -41,15 +42,22 @@ genrule(
 
 java_library(
     name = "junit5-compile",
-    srcs = glob(["*.java"]),
+    srcs = glob(
```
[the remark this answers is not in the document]
fmeum: Could you explain why `--release 8` isn't sufficient here? It should equally prevent builds from breaking if `SecurityManager` is removed in the future, but different from `--release 11` allows compilation with Java 8.

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BUILD.bazel
```diff
@@ -62,6 +70,15 @@ java_library(
     ],
 )
 
+# Do not build with --release 8 to work around https://bugs.openjdk.org/browse/JDK-8214165.
+java_library(
```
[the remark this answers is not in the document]
fmeum: These classes specifically don't refer to `SecurityManager` (in fact, they only use `java.lang.reflect.Field`, `java.util.Map`, and `sun.misc.Unsafe`). Is there another reason to specify a fixed `--release`? 

Specifying `--release 11` would break JDK 8 compatibility and specifying `--release 8` isn't possible due to the linked JDK bug.

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BUILD.bazel
fmeum: This adds the transitive deps to the compilation classpath, but if consumers depend on them strict deps checks will catch that and fail the build. `neverlink` ensures that the transitive deps will not be on the runtime classpath.

The only difference I can see is that this could allow tests to build if they accidentally depend on the transitive deps of this target *and* a non-default value of `--strict_java_deps` is used. But even in that case, running the test would still fail since the transitive deps are not on the runtime classpath.

Let me know if you still prefer the old approach and I can revert back to it.

fmeum:
The test pipeline failure looks like a flake. I pushed a commit that should fix the format pipeline failure.

fmeum:
@shs96c Friendly ping, resolved the conflict.

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/Java17SystemExitToggle.java
shs96c: Just tried this with Java 19 and it appears to work as expected.

shs96c (changes requested):
There's a compilation failure caused by the use of `List.of` in the `ActualRunner`, and I'd like us to have no risk of leaking our deps into someone else's classpath if we can help it. Once the code compiles, we can merge, and if necessary I can tweak the build files.

Apologies for the delay in this review. Work got on top of me, and I didn't give this the attention it needed.

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BUILD.bazel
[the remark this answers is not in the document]
fmeum: I expanded on why we can't in the comment: `--release 8` breaks due to a JDK bug, `--release N` with N > 8 obviously means we aren't compatible with Java 8 anymore.

The source files in this target reference only very few classes, all of which are extremely unlikely to go away before Java 8 itself, so I don't think it's a problem to not specify `--release` on this target.

fmeum (commented):
Rebased onto main and fixed all the remaining Java 8 incompatibilities. I also restored the original approach to avoid deps pollution.

shs96c (approved):
Thank you for sticking with this! LGTM

PR #135, opened 2023-02-08
More support for java test suites.

tjoneslo:
This makes two changes to support further use of the Java Test suites matching what the junit5 testing allows. To that end there are two small changes here:
- expose the create_jvm_test_suite() macro to have access to set the define_test() for duplicate name processing.
- set the create_lint_tests() macro to avoid creating duplicate linter tests.

### DIFF
--- a/java/private/library.bzl
+++ b/java/private/library.bzl
@@ -1,4 +1,5 @@
 load("@apple_rules_lint//lint:defs.bzl", "get_lint_config")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 load("@rules_jvm_external//:defs.bzl", _java_export = "java_export")
 load("//java/private:checkstyle.bzl", "checkstyle_test")
 load("//java/private:pmd.bzl", "pmd_test")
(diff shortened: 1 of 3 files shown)


shs96c (approved):

java/private/library.bzl
```diff
@@ -37,7 +37,7 @@ def create_lint_tests(name, **kwargs):
         )
 
     spotbugs = get_lint_config("java-spotbugs", tags)
-    if spotbugs != None:
+    if spotbugs != None and not native.existing_rule("%s-spotbuts" % name):
```
illicitonion, line 40:
```suggestion
    if spotbugs != None and not native.existing_rule("%s-spotbugs" % name):
```

illicitonion: Would it make more sense to use a `maybe` rather than hard-coding its impl here in ways which could contain typos?

tjoneslo: `maybe()` is a repository rule, and this isn't loading a repository. Good thought though.

illicitonion: There's actually nothing repository-rule specific about `maybe`, it works everywhere!

illicitonion (approved):
Thanks!

PR #137, opened 2023-02-10
release: publish release archive

alexeagle:
GitHub's stability guarantee for the archive is iffy, and we want metrics on downloads.

See https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1411222578


alexeagle:
@shs96c looks like the CI is red at HEAD, who should I ping about that?

illicitonion:
> @shs96c looks like the CI is red at HEAD, who should I ping about that?

I'm looking at it as we speak :)

illicitonion:
Resolved in https://github.com/bazel-contrib/rules_jvm/pull/138

illicitonion (approved):

PR #139, opened 2023-02-20
Identify fully-qualified types

illicitonion:
This adds three contexts where we look for fully qualified types:
* Fully-qualified static method calls
* Fully-qualified constructor calls
* Variable assignments with fully-qualified types

It's a bit heuristic - it guesses what's a type based on standard conventions of class names.


shs96c (approved):

shs96c:
LGTM. The logic around figuring out whether something is a class or not makes me wince, but we can iterate on it.

PR #141, opened 2023-02-22
Blank out line numbers in test expectations

illicitonion:
Otherwise adding/removing lines to production code changes test expectations.

PR #142, opened 2023-02-22
Use sorted set instead of manual impl

illicitonion:
The old implementation round-tripped labels through a label -> string -> label -> string conversion loop, and had manual deduplicating and sorting.

### DIFF
--- a/java/gazelle/resolve.go
+++ b/java/gazelle/resolve.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 
 	"github.com/bazel-contrib/rules_jvm/java/gazelle/private/java"
+	"github.com/bazel-contrib/rules_jvm/java/gazelle/private/sorted_set"
 	"github.com/bazelbuild/bazel-gazelle/config"
 	"github.com/bazelbuild/bazel-gazelle/label"
 	"github.com/bazelbuild/bazel-gazelle/repo"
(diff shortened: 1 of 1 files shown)


illicitonion:
(Test failure will be fixed by https://github.com/bazel-contrib/rules_jvm/pull/141 :))

java/gazelle/resolve.go
```diff
@@ -105,48 +103,34 @@ func (jr Resolver) Resolve(c *config.Config, ix *resolve.RuleIndex, rc *repo.Rem
...
 }
 
+func labelLess(l, r label.Label) bool {
+	if l.Relative {
+		if r.Relative {
+			return l.String() < r.String()
+		}
+		return true
+	}
+	if r.Relative {
```
shs96c, line 128: It's weird, but couldn't you have someone use the fully qualified label to refer to something in the same package? And doing so would fail this test.

java/gazelle/resolve.go
```diff
@@ -105,48 +103,34 @@ func (jr Resolver) Resolve(c *config.Config, ix *resolve.RuleIndex, rc *repo.Rem
...
+		return false
+	}
+	return l.String() < r.String()
```
shs96c, line 131: Is there a way to get the fully-qualified label? If there is, then this would be enough of a check

java/gazelle/resolve.go
illicitonion: That's fine here, this is just for sorting, where there isn't really a concept of "the current package" - if people want to canonicalise labels so that in-package ones are relative, they should do so before trying to sort.

java/gazelle/resolve.go
illicitonion: That's exactly what `.String()` does on a `Label` :)

shs96c: Then isn't this final check enough, so you don't need the additional checks above? That is, this entire method could be boiled down to this final line?

shs96c (approved):
LGTM, with one open question

java/gazelle/resolve.go
illicitonion: Not quite - `/` sorts before `:` in UTF-8, so just sorting by stringification would put absolute labels before relative ones. Added a comment to the code explaining why we do this.

PR #145, opened 2023-02-22
javaparser returns classes and packages separately

illicitonion:
Previously, some of the values in imports were classes and others were
packages. This makes handling them on the client side more fiddly.
Instead, differentiate between the two where we know which we're
handling.

Also, fix up some tests and documentation which were obsolete,
clarifying that we don't strip standard library types here any more,
this is now done on the client side.

I'm about to introduce much stronger typed handling of classes vs
packages in the Go code, so the ugly "double-call addNonLocalImports"
pattern will go away shortly.

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/javaparser/generators/ClasspathParser.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/javaparser/generators/ClasspathParser.java
@@ -44,6 +44,7 @@ public class ClasspathParser {
   private static final Logger logger = LoggerFactory.getLogger(ClasspathParser.class);
 
   private final Set<String> usedTypes = new TreeSet<>();
+  private final Set<String> usedPackagesWithoutSpecificTypes = new TreeSet<>();
   private final Set<String> packages = new TreeSet<>();
   private final Set<String> mainClasses = new TreeSet<>();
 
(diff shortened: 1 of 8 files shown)


java/src/com/github/bazel_contrib/contrib_rules_jvm/javaparser/generators/ClasspathParser.java
```diff
@@ -63,6 +64,10 @@ public Set<String> getUsedTypes() {
     return usedTypes;
   }
 
+  public Set<String> getUsedPackagesWithoutSpecificTypes() {
+    return usedPackagesWithoutSpecificTypes;
```
shs96c, line 68: For future reference, because you're returning the actual field and that's marked as being immutable, anyone calling this method can now add or remove items from the returned value. A defensive copy to `ImmutableSet` (which preserves insertion order) is normally a Good Idea.

shs96c (approved):
LGTM. I'll leave this to you to merge.

java/src/com/github/bazel_contrib/contrib_rules_jvm/javaparser/generators/ClasspathParser.java
illicitonion: Good call - added the defensive copies (and changed the signatures to return `Immutable*`, which I know isn't strictly necessary, but I generally find nice to do :))

PR #146, opened 2023-02-22
Convert junit5 tests to be pure junit5

illicitonion:
Right now we're detecting both junit4 and junit5 features of these tests, because we're using Assert from junit4.

Instead, make these tests pure junit5. We have specific tests for testing junit4/junit5/mixed behaviour.


shs96c (approved):
LGTM

PR #153, opened 2023-02-27
Test code more accurately detects dependencies

### LINKED ISSUE #136: [Gazelle] Resolve deps that are testonly libraries in the test hierarchy
I added a Spring Boot test configuration that is meant to be shared among many tests.  Since it is a library that is meant only to be used in tests, I added it in the `/src/test/` subtree of my project.  I ran gazelle and it generated a `junit5_test_suite`, but since it is a library,  I adjusted the generated BUILD rule to be a `java_library` (and also set the visibility and testonly flags appropriately).  However, gazelle was not able to pick up this library and automatically add it as a dependency.  I was able to work around this by adding a `gazelle:resolve` directive, but it would be nice if that wasn't the case.

illicitonion:
In particular, two classes of dependency are now better handled:
1. Test code for which there isn't a corresponding non-test package no longer emits a warning about a missing package.
2. Test code can now depend on test utility helpers which are defined in other test packages, in very specific circumstances.

I strongly recommend reviewing commit-by-commit.

Fixes #136


java/gazelle/resolve_test.go
```diff
@@ -63,10 +65,10 @@ java_library(
...
     name = "myproject",
     srcs = ["App.java"],
-    _imports = [
-        "com.google.common.primitives.Ints",
-        "java.lang.Exception",
-        "java.lang.String",
+    _imported_packages = [
+        "com.google.common.primitives",
+        "java.lang",
+        "java.lang",
```
shs96c, line 72: Query: I thought this was a set, not a list.

java/gazelle/private/types/types.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func PackageNameLess(l, r PackageName) bool {
+	return l.Name < r.Name
+}
+
+type ClassName struct {
+	packageName        PackageName
+	bareOuterClassName string
+	innerClassNames    []string
```
shs96c, line 32: Classes can be nested to an arbitrary depth. I can't tell if `bareOuterClassName` is the containing class or the top-level class

java/gazelle/private/types/types.go
```diff
@@ -0,0 +1,85 @@
...
+
+func NewClassName(packageName PackageName, bareOuterClassName string) ClassName {
+	return ClassName{
+		packageName:        packageName,
+		bareOuterClassName: bareOuterClassName,
+	}
+}
+
+func ParseClassName(fullyQualified string) (*ClassName, error) {
+	parts := strings.Split(fullyQualified, ".")
```
shs96c, line 61: I guess we've already handled the case where in Java an inner class is represented as something like `com.example.Foo$Bar`? We can rely on the fully qualified name just having period characters separating things?

shs96c (approved):
Seems fine, but I have questions. They're not serious enough to block landing the PR, though.

java/gazelle/resolve_test.go
illicitonion: Good point! Removed (accidental) duplication.

java/gazelle/private/types/types.go
illicitonion: Yeah, we're only getting these values from source code itself, where $s aren't present; we don't currently parse e.g. filenames from inside jar files.

PR #156, opened 2023-02-28
Share default test patterns with java_test_suite

illicitonion:
And allow passing arbitrary regexes.


java/gazelle/javaconfig/codegen/codegen.go
```diff
@@ -0,0 +1,74 @@
...
+							} else {
+								log.Fatalf("DEFAULT_TEST_SUFFIXES values must be strings but got %v", reflect.TypeOf(expr))
+							}
+						}
+					} else {
+						log.Fatalf("DEFAULT_TEST_SUFFIXES must be a list but got %v", reflect.TypeOf(assign.RHS).Name())
+					}
+				}
+			}
+		}
```
stevebarrau, line 54: Can we dedent this loop using early return and continue, aka [left edge happy path](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88) technique?

java/gazelle/configure.go
```diff
@@ -55,13 +55,18 @@ func (jc *Configurer) KnownDirectives() []string {
 		javaconfig.ModuleGranularityDirective,
 		javaconfig.TestMode,
 		javaconfig.ExcludeArtifact,
+		javaconfig.TestFilePatterns,
```
stevebarrau, line 58: Can we sort this list?

java/gazelle/configure.go
```diff
@@ -102,6 +107,8 @@ func (jc *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 				cfg.SetTestMode(d.Value)
 			case javaconfig.ExcludeArtifact:
 				cfg.AddExcludedArtifact(d.Value)
+			case javaconfig.TestFilePatterns:
+				cfg.SetTestFilePatterns(d.Value)
```
stevebarrau, line 111: Can we keep this switch case sorted? And with the same whitespace line after the cases.

java/gazelle/javaconfig/BUILD.bazel
```diff
@@ -1,11 +1,31 @@
...
     deps = [
         "@com_github_bazelbuild_buildtools//build:go_default_library",
     ],
 )
+
+genrule(
+    name = "generate_default_java_test_patterns_src",
+    srcs = ["//java/private:java_test_suite_shared_constants.bzl"],
+    outs = ["default_test_patterns.go"],
+    cmd = "$(execpath //java/gazelle/javaconfig/codegen) $< $@",
```
stevebarrau, line 20: What is the difference between `execpath` and `location`?

java/gazelle/javaconfig/codegen/codegen.go
```diff
@@ -0,0 +1,74 @@
...
+					if rhs, ok := assign.RHS.(*bzl.ListExpr); ok {
+						for _, expr := range rhs.List {
+							if lit, ok := expr.(*bzl.StringExpr); ok {
+								for _, r := range []rune(lit.Value) {
+									if !unicode.IsLetter(r) && !unicode.IsNumber(r) && r != '.' {
+										// We're going to convert these strings to regexes - this is a simple check to make sure we can do so safely.
+										log.Fatalf("DEFAULT_TEST_SUFFIXES may only contain alphanumeric characters or . but got %q", lit.Value)
+									}
+								}
+								defaultTestSuffixes = append(defaultTestSuffixes, lit.Value)
```
stevebarrau, line 44: Should we fail if the list is re-defined instead of doing an append?

In the input is:

```starlark
DEFAULT_TEST_SUFFIXES = ["Test.java"]
DEFAULT_TEST_SUFFIXES = ["Test.java"]
DEFAULT_TEST_SUFFIXES = ["Test.java"]
```

The current output is:

```go
package javaconfig

import "regexp"

var defaultTestRegexps = []*regexp.Regexp{
regexp.MustCompile(".*Test\\.java$"),
regexp.MustCompile(".*Test\\.java$"),
regexp.MustCompile(".*Test\\.java$"),
}
```

java/gazelle/javaconfig/codegen/codegen.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	output := `package javaconfig
+
+import "regexp"
+
+var defaultTestRegexps = []*regexp.Regexp{
+`
+
+	for _, dts := range defaultTestSuffixes {
+		output += "regexp.MustCompile(\".*" + strings.ReplaceAll(dts, ".", "\\\\.") + "$\"),\n"
```
stevebarrau, line 67: This does not have to generate formatted code, but could we prefix these lines with 1 tab?

java/gazelle/javaconfig/codegen/codegen.go
illicitonion: Done - I always get seduced by the `if foo, ok := bar; ok` syntax, and then get sad that it leads to all the rightward drift

java/gazelle/configure.go
illicitonion: Done - also renamed all the constants to have a `Java` prefix so that the constants and the values sort the same.

java/gazelle/configure.go
illicitonion: Done

java/gazelle/javaconfig/BUILD.bazel
illicitonion: See https://bazel.build/reference/be/make-variables#predefined_label_variables - `location` is a legacy thing which expands to one of `execpath` or `rootpath` depending on where it's used.

java/gazelle/javaconfig/codegen/codegen.go
illicitonion: Absolutely! Done.

java/gazelle/javaconfig/codegen/codegen.go
illicitonion: Done

java/gazelle/javaconfig/codegen/BUILD.bazel
```diff
@@ -0,0 +1,15 @@
...
+    srcs = ["codegen.go"],
+    importpath = "github.com/bazel-contrib/rules_jvm/java/gazelle/javaconfig/codegen",
+    visibility = ["//visibility:private"],
+    deps = ["@net_starlark_go//starlark"],
+)
+
+go_binary(
+    name = "codegen",
+    embed = [":codegen_lib"],
+    visibility = ["//visibility:public"],
```
stevebarrau, line 14: This may be fine tuned

stevebarrau (approved):
Thanks for finding out about this discrepancy.

Issue #154, opened 2023-02-27
Should gazelle be removing runtime_deps?


caseyduquettesc:
I have noticed that running Gazelle removes any runtime deps that are missing `# keep`. I'm wondering if that's really desired behavior because presumably Gazelle couldn't know what deps a library needs at runtime? Should Gazelle be ignoring that attribute, so we don't need to add `# keeps` everywhere?

For the sake of the discussion, here's an example

```bzl
load("@rules_java//java:defs.bzl", "java_library")

java_library(
    name = "client",
    visibility = ["//:__subpackages__"],
    runtime_deps = ["@maven//:org_glassfish_jersey_core_jersey_common"],
    deps = ["@maven//:javax_ws_rs_javax_ws_rs_api"],
)
```

Gazelle will always remove `@maven//:org_glassfish_jersey_core_jersey_common` here because it isn't aware that `InternalServerErrorException` tries to dynamically load a class.

<img width="754" alt="image" src="https://user-images.githubusercontent.com/16676170/221700217-e5ecd7d1-4f81-40e6-9757-96253112154f.png">

<img width="873" alt="image" src="https://user-images.githubusercontent.com/16676170/221700743-95222568-57a4-46f9-b53c-a24da54676a7.png">

Was this a deliberate design decision that I'm just not seeing the reason for?

stevebarrau:
Thanks for the ticket @caseyduquettesc. We came to the same conclusion recently. This was not a deliberate design decision, I'll work on a fix for this.

PR #157, opened 2023-03-01
feat(gazelle): Keep existing runtime_deps

stevebarrau:
Keep existing runtime_deps on java_* rules.

In this new style, the gazelle extension will only add to the runtime_deps attribute, never remove.

This only works on []string runtime_deps. More complex types would run into the Gazelle mergeability issues anyways.

Closes https://github.com/bazel-contrib/rules_jvm/issues/154


java/gazelle/generate.go
```diff
@@ -346,6 +348,36 @@ func (l javaLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
...
+	if file == nil {
+		return runtimeDeps
+	}
+
+	for _, r := range file.Rules {
+		if r.Kind() != kind || r.Name() != name {
+			continue
+		}
+
+		// This does not support non string list values from runtime_deps.
```
illicitonion, line 362:
```suggestion
		// This does not support non string list values from runtime_deps.
		// Currently, that means if a target has a runtime_deps of a different kind (e.g. a select), we will remove it.
		// Hopefully in the future we can be less destructive.
```

illicitonion (approved):
LGTM, thanks!

java/gazelle/generate.go
stevebarrau: Add in b040e7311788ff875faf389229864421af2f5843, with an example that shows this behavior.

java/gazelle/testdata/lib_with_complex_runtime_deps/README.md
```diff
@@ -0,0 +1,5 @@
+# Complex runtime_deps
+
+Loading existing runtime_deps attribute only supports string list values.
+
+Other type are ignored and deleted.
```
illicitonion, line 5:
```suggestion
Other type are ignored and deleted.

Note that this is not the desired behaviour, simply documenting the current behaviour. Improving this would be welcome.
```

illicitonion (approved):

PR #158, opened 2023-03-07
gazelle: Generate exports for non-private return types

illicitonion:
Return values which are used by calling code require the return type to be on the classpath in order to compile uses of that value. Accordingly, we should export those dependencies, to make sure the depender has them on the classpath while compiling, to avoid strict deps issues.

We don't do this for private functions, but all other visibilities may be called from other targets, so may need exporting.


java/gazelle/README.md
```diff
@@ -81,6 +81,11 @@ Currently, the gazelle plugin makes the following assumptions about the code it'
 1. Non-test code doesn't depend on test code.
 1. Non-test code used by one package of tests either lives in the same directory as those tests, or lives in a non-test-code directory. We also detect non-test code used from another test package, if that other package doesn't have a corresponding non-test code directory, but require you to manually set the visibility on the depended-on target, because this is an unexpected set-up.
 1. Package names and class/interface names follow standard java conventions; that is: package names are all lower-case, and class and interface names start with Upper Case letters.
+1. Code doesn't use types which it doesn't name _only_ through unnamed means, across multiple calls. For example, if some code calls `x.foo().bar()` where the return type of `foo` is defined in another target, and the calling code explicitly uses a type from that target somewhere else. In the case of `x.foo()`, we add exports so that the caller will have access to the return type of `foo()`, but we assume that if you call a chain of _multiple_ calls, your code will have some other dependency on the return types of the subsequent return types.
```
blorente, line 84:
```suggestion
1. Code doesn't use types which it doesn't name _only_ through unnamed means, across multiple calls. For example, if some code calls `x.foo().bar()` where the return type of `foo` is defined in another target, and the calling code explicitly uses a type from that target somewhere else. In the case of `x.foo()`, we add exports so that the caller will have access to the return type of `foo()`, but do not track dependencies on the return types across _multiple_ calls.
```

java/src/com/github/bazel_contrib/contrib_rules_jvm/javaparser/generators/ClasspathParser.java
```diff
@@ -284,23 +295,39 @@ public Void visitVariable(VariableTree node, Void unused) {
...
+          types.add(importedType);
+        } else if (components.size() > 1) {
           usedTypes.add(typeName);
+          types.add(typeName);
+        } else {
+          String expandedImport = currentFileImports.get(typeName);
+          if (expandedImport != null) {
+            usedTypes.add(expandedImport);
+            types.add(expandedImport);
+          }
```
blorente, line 317: I don't know when this branch will trigger. If I'm reading this correctly, we have:
- `if (currentFileImports.containsKey(components.get(0))) {` -> Will trigger for both `IDENTIFIER` and `MEMBER_SELECT` where we've already imported the left hand side.
- ` else if (components.size() > 1) {` -> Unconditionally import all other `MEMBER_SELECT`s, because they're by definition not in the currently used imports.

It feels like the only types left for this `else` clause are `IDENTIFIER`s that are _not_ already in `currentFileImports`, so this `get` will always return `null`.

java/src/com/github/bazel_contrib/contrib_rules_jvm/javaparser/generators/ClasspathParser.java
```diff
@@ -284,23 +295,39 @@ public Void visitVariable(VariableTree node, Void unused) {
...
         }
       } else if (identifier.getKind() == Tree.Kind.PARAMETERIZED_TYPE) {
         Tree baseType = ((ParameterizedTypeTree) identifier).getType();
         checkFullyQualifiedType(baseType);
         ((ParameterizedTypeTree) identifier)
-            .getTypeArguments()
-            .forEach(this::checkFullyQualifiedType);
+            .getTypeArguments().stream()
+                .flatMap(t -> checkFullyQualifiedType(t).stream())
+                .forEach(types::add);
```
blorente, line 325: TIL member functions support the `::` syntax :D 
This is fine code, but something in me just assumed that `types::add` would be a static function. Maybe it was the `::`.

blorente (commented):
Just looked at the java side, and LGTM.

java/src/com/github/bazel_contrib/contrib_rules_jvm/javaparser/generators/ClasspathParser.java
illicitonion: You're right, I don't think it can happen - removed

java/gazelle/resolve.go
```diff
@@ -80,25 +80,26 @@ func (Resolver) Embeds(r *rule.Rule, from label.Label) []label.Label {
...
 	resolveInput := imports.(types.ResolveInput)
-	javaImports := resolveInput.ImportedPackageNames
-	if javaImports.Len() == 0 {
-		return
-	}
 
-	deps := sorted_set.NewSortedSetFn[label.Label]([]label.Label{}, labelLess)
+	isTestRule := isTestRule(r.Kind())
+
+	jr.populateAttr(c, r, "deps", resolveInput.ImportedPackageNames, ix, isTestRule, from, resolveInput.PackageNames)
```
shs96c, line 86: What about the `runtime_deps`?

shs96c (approved):

java/gazelle/resolve.go
illicitonion: Great question! See this comment: https://github.com/bazel-contrib/rules_jvm/blob/9586f4edb573ab9b63569cc1ac6aecc532977bd2/java/gazelle/generate.go#L461-L464

(tl;dr: It's an existing TODO for the future)

PR #161, opened 2023-03-10
Ensure that sysout and syserr are escaped

shs96c:
Sadly, the IJ console won't unescape the content, but this is a definite step forwards.


illicitonion (approved):
Thanks!

PR #162, opened 2023-03-13
Add test sharding support to the JUnit 5 runner

fmeum:
Since JUnit 5 test plans can have complicated structures, the sharding algorithm uses a hash of the test descriptor's unique ID to determine which shard to assign a given test to.

Dynamic tests and test templates are not affected by PostDiscoveryFilter and thus not individually sharded.


shs96c (commented):
Looks great. Thank you! It would be nice if there were some tests, though perhaps that'd be hard to do.

fmeum:
I went for an integration test instead, please take a look

shs96c (approved):
Thanks for adding the tests! LGTM

PR #164, opened 2023-03-14
Use TestExecutionListener in shard test

fmeum:
Makes the test slightly more realistic and is needed for future work on sharding parameterized tests.


fmeum:
@shs96c Small follow-up while I try to get parameterized test sharding working.

shs96c (commented):
I am surprised that adding tests dynamically like this works, but if you've tested and verified, this LGTM

shs96c (approved):
LGTM

PR #165, opened 2023-03-16
Expand on the ability to include the same test in two suites

tjoneslo:
This is to expand the ability of the user to create their own "define_test()" to allow non-default behavior for creating tests within the suite. The goal for this was to add the same test class to two different suites and run them with different configurations.

Add documentation about expected use of the define_test() macro


illicitonion (approved):
LGTM - thanks!

PR #175, opened 2023-04-18
junit5: Print failure causes

fmeum:
When the command-line summary prints a filtered stack trace for a failure, it now also prints similarly filtered stack traces for causes.


shs96c (approved):
LGTM. Makes sense, and better support for actually finding what went wrong is very helpful. Thank you!

Issue #176, opened 2023-04-20
Please help bump PMD version


rdesgroppes:
Currently shipped PMD version is at 6.46.0, released in May 2022.
Since then, a handful of fixes & enhancements got released, among which:
- https://github.com/pmd/pmd/issues/3936 in 6.47.0
- https://github.com/pmd/pmd/issues/4015 in 6.48.0
- https://github.com/pmd/pmd/issues/4082 in 6.49.0
- https://github.com/pmd/pmd/issues/4085 in 6.50.0
- https://github.com/pmd/pmd/issues/4139 in 6.51.0
- https://github.com/pmd/pmd/issues/4152 in 6.52.0
- https://github.com/pmd/pmd/issues/4266 in 6.53.0
- https://github.com/pmd/pmd/issues/4364 in 6.54.0
- https://github.com/pmd/pmd/issues/4333 in 6.55.0

The latter also [states](https://docs.pmd-code.org/pmd-doc-6.55.0/pmd_release_notes.html#pmd-7-development) that:
> This release is the last planned release of PMD 6. The first version 6.0.0 was released in December 2017. Over the course of more than 5 years we published almost every month a new minor version of PMD 6 with new features and improvements.

Before PMD 7 is out, I therefore gave a try bumping up PMD version to 6.55.0 but miserably failed running `tools/update-dependencies.sh`, which fails on my machine:
```
$ ./tools/update-dependencies.sh
go: downloading github.com/rs/zerolog v1.26.1
[...]
ERROR: Skipping '@unpinned_maven//:pin': No repository visible as '@unpinned_maven' from main repository
WARNING: Target pattern parsing failed.
ERROR: No repository visible as '@unpinned_maven' from main repository
INFO: Elapsed time: 0.096s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
ERROR: Build failed. Not running target
```

Could you please help (or do) bump PMD version to 6.55.0?

rdesgroppes:
@shs96c Any idea? Thanks in advance!

rdesgroppes:
Update: seems #115 didn't account for the presence of `maven` in `tools/update-dependencies.sh`.

PR #170, opened 2023-03-24
Add a bzlmod module definition

shs96c:

PR #177, opened 2023-04-26
Fix `tools/update-dependencies.sh`

rdesgroppes:
On the one hand, #115 didn't account for the presence of `maven` in `tools/update-dependencies.sh`. On the other hand, it turns out that re-pinning dependencies requires that `bzlmod` be disabled, as [was done](https://github.com/shs96c/rules_jvm/blob/cb56c56184e6358d95a8f4a546b524edd14f5cca/tools/update-docs.sh#L4) for `stardoc` in #170.

The present change therefore addresses these issues and applies resulting updates.

This should unblock #176.


illicitonion (approved):
Thanks!

rdesgroppes:
> Thanks!

Wow, that was fast, I'm the one to thank you!

PR #179, opened 2023-05-05
pass tag for java_test_lib

lyao-77:
when we add `no-tag` to ignore linter checks for all test code, it only applies to `test` target. We need to make the expanded `java_test_lib` target inherit this behavior as well.


lyao-77:
Hi @illicitonion, can I get a review and approval for this small change? Thanks

illicitonion (approved):
LGTM - thanks!

PR #183, opened 2023-07-11
build: Upgrade grpc-java to 1.56.1 and protobuf to 21.7 and ensure we choose the right version

blorente:
And add it as an explicit dependency of the gazelle plugin.
Otherwise, a random one would be picked, usually from one of the grpc dependencies.


illicitonion (approved):
Thanks!

PR #190, opened 2023-07-21
Handle excluded artifacts defined in non-root packages

Ignas:
This handles different sub-packages having different sets of excluded artifacts, also fixes a crash when such a case occurs.


illicitonion (approved):
This looks great, thanks so much!

illicitonion:
I've just kicked off release v0.17.0 which should include this change - thanks again for the contribution!

PR #192, opened 2023-07-25
Added java_maven_repository_name directive

Ignas:
Some repositories can run maven_install with a different name or maintain multiple sets of maven dependencies, this change will allow overriding `@maven` with whatever custom name that is needed.


illicitonion (approved):
Thanks so much!

PR #195, opened 2023-08-15
junit5: pass static nested classes to the launcher

gaurav-narula:
This change iterates over all the static nested classes of the test class and adds them to LauncherDiscoveryRequestBuilder to allow running tests within static nested classes.

While Java allows more than one level of nesting, we limit our search to one level of nesting at the moment.

Junit4 tests annotated with `RunWith(Enclosed.class)` need to be special cased.

Refer: https://github.com/gradle/gradle/issues/4427


shs96c (approved):
LGTM. Thank you for the PR, and nice catch!

PR #199, opened 2023-08-30
Handle java_library being remapped with map_kind

Ignas:
When preserving runtime dependencies of existing rules use the new rule name if it was remapped.


illicitonion (approved):
Thanks!

PR #200, opened 2023-08-31
Remove empty glob

### LINKED ISSUE #198: Failure with --incompatible_disallow_empty_glob
The `data = glob(["testdata/**"])` in https://github.com/bazel-contrib/rules_jvm/blob/fae67e4df87decc0c16902230d50539a2e1897bc/java/gazelle/BUILD.bazel#L50 is empty because the `testdata` subdirectory contains a `BUILD.bazel` file and globs don't cross package boundaries. This causes the build to fail when `--incompatible_disallow_empty_glob` is set.

This will eventually become the default (tracked in https://github.com/bazelbuild/bazel/issues/8195 )

illicitonion:
This directory is a package.

Fixes #198.

### DIFF
--- a/java/gazelle/BUILD.bazel
+++ b/java/gazelle/BUILD.bazel
@@ -47,7 +47,6 @@ go_test(
         "generate_test.go",
         "resolve_test.go",
     ],
-    data = glob(["testdata/**"]),
     embed = [":gazelle"],
     deps = [
         "//java/gazelle/javaconfig",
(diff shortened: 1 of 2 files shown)


java/gazelle/BUILD.bazel
```diff
@@ -47,7 +47,6 @@ go_test(
         "generate_test.go",
         "resolve_test.go",
     ],
-    data = glob(["testdata/**"]),
```
stevebarrau, line 50: Please add the `# gazelle:exclude testdata` gazelle [directive](https://github.com/bazelbuild/bazel-gazelle#directives) to this file; this prevent running format.sh from adding this back.

stevebarrau (changes requested):

stevebarrau (approved):

PR #202, opened 2023-09-04
gazelle: Fail when errors occur

stevebarrau:
Instead of introducing a fail-fast, leverage the lifecycle to fail the extension when errors that would produce invalid build files occur.


illicitonion (commented):
LGTM! Can you add an integration test or two showing this working?

`java/gazelle/testdata/lib_with_invalid_label_in_runtime_deps` has an example of a test where we run gazelle and expect it to fail with specific output.

stevebarrau:
Added 2 tests:

- one where there is a collision
- one where there is a collision and the developer added the required resolve directive

illicitonion (approved):
Thanks!

PR #203, opened 2023-09-04
test(gazelle): Use json instead of color logs

stevebarrau:
Rely on json formatted logs to avoid post-precessing and unstable color format.


java/gazelle/testdata/test_without_production_code/BUILD.out
```diff
@@ -0,0 +1 @@
+# gazelle:resolve java org.junit @maven//:junit_junit
```
illicitonion, line 1: What the story with this addition?

illicitonion (approved):
Much cleaner, thanks!

java/gazelle/testdata/test_without_production_code/BUILD.out
stevebarrau: I do not recall it. Removed in the rebase

PR #217, opened 2023-10-30
junit5: add property to allow SecurityManager override in tests

### LINKED ISSUE #213: Allow setting SecurityManager
We have a need to allow unit tests set a `SecurityManager` as part of their testing. As such we would like to contribute a way to prevent installation of the default `SecurityManager`. 

Can we get some guidance on the wanted/acceptable design?

**A) Using Test Runner**

It looks like one possible way is to introduce [a new test runner](https://github.com/bazel-contrib/rules_jvm/blob/e384e30182b3f36ef2f324943e06567c7577af79/java/private/create_jvm_test_suite.bzl#L11):
```
_RUNNERS = [
    "junit4",
    "junit5",
    "junit5_without_security_manager",
]
```

This test runner would use a [custom test definition function](https://github.com/bazel-contrib/rules_jvm/blob/e384e30182b3f36ef2f324943e06567c7577af79/java/private/java_test_suite.bzl#L23C5-L23C24):
```
_TEST_GENERATORS = {
    "junit4": _define_junit4_test,
    "junit5": _define_junit5_test,
    "junit5_without_security_manager": _define_junit5_test_without_security_manager,
}

def _define_junit5_test_without_security_manager(name, **kwargs):
    java_junit5_test(
        name = name,
        include_engines = kwargs.pop("include_engines", None),
        exclude_engines = kwargs.pop("exclude_engines", None),
        disable_security_manager = True,
        **kwargs
    )
    return name
```

And then the logic in [junit5.bzl](https://github.com/bazel-contrib/rules_jvm/blob/e384e30182b3f36ef2f324943e06567c7577af79/java/private/junit5.bzl#L25) to:
* add attribute `disable_security_manager`
* skip any SecurityManager work if `disable_security_manager` is set to `True`
* use a different [main_class](https://github.com/bazel-contrib/rules_jvm/blob/e384e30182b3f36ef2f324943e06567c7577af79/java/private/junit5.bzl#L101C9-L101C19) if `disable_security_manager` is set to `True`

As well as a new main class [based on JUnit5Runner](https://github.com/bazel-contrib/rules_jvm/blob/e384e30182b3f36ef2f324943e06567c7577af79/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/JUnit5Runner.java#L24) that skips all the system exit stuff. 

**B) Using a special system property**

The junit5 macro already [has special support for detecting **-Djava.security.manager=** jvm_flag](https://github.com/bazel-contrib/rules_jvm/blob/e384e30182b3f36ef2f324943e06567c7577af79/java/private/junit5.bzl#L94).

Can we add another `jvm_flag` (eg., `-Dbazel.junit5runner.skipSecurityManagerInstallation=true`) that once set will:
* trigger the macro to not set any `-Djava.security.manager=` options
* trigger existing `Junit5Runner` to not install any SecurityManager

Option B seems easier to implement and would make it less obvious that this mode of operation is supported.

It might be possible to implement option B *without* a new system property. For example, when `-Djava.security.manager=my.custom.Class` is set we could change the `JUnit5Runner` to not install its SecurityManager at all. However, that would be a breaking change in terms of behavior for existing users *and* might not work for us because the code under tests needs to call `setSecurityManager`.

rsalvador:
Fixes #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,6 +316,8 @@ you are using the standard `@maven` namespace for your
 **Note**: The junit5 runner prevents `System.exit` being called
 using a `SecurityManager`, which means that one test can't
 prematurely cause an entire test run to finish unexpectedly.
+This security measure prohibits tests from setting their own `SecurityManager`.
+To override this, set the `bazel.junit5runner.allowSettingSecurityManager` system property.
 
 While the `SecurityManager` has been deprecated in recent Java
 releases, there's no replacement yet. JEP 411 has this as one of
--- a/java/private/junit5.bzl
+++ b/java/private/junit5.bzl
@@ -57,6 +57,8 @@ def java_junit5_test(
     **Note**: The junit5 runner prevents `System.exit` being called
     using a `SecurityManager`, which means that one test can't
     prematurely cause an entire test run to finish unexpectedly.
+    This security measure prohibits tests from setting their own `SecurityManager`.
+    To override this, set the `bazel.junit5runner.allowSettingSecurityManager` system property.
 
     While the `SecurityManager` has been deprecated in recent Java
     releases, there's no replacement yet. JEP 411 has this as one of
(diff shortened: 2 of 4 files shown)


guw:
@illicitonion @shs96c Can we can this reviewed? 🙏

shs96c (approved):
Seems sensible to me. Thank you for creating this PR!

PR #221, opened 2023-11-09
junit5: Add support for `--test_runner_fail_fast`

fmeum:
When the Bazel flag `--test_runner_fail_fast` is enabled, the JUnit 5 runner will now skip all other tests after the first one has failed in an attempt to speed up the (failing) test run. This is useful in some CI setups as well as during bisection.

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BUILD.bazel
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BUILD.bazel
@@ -58,11 +58,18 @@ java_library(
         "--release",
         "8",
     ],
+    resource_strip_prefix = package_name(),
+    resources = ["META-INF/services/org.junit.jupiter.api.extension.Extension"],
     deps = [
         ":system-exit-toggle",
         # The only dependencies here are those required to run
         # a junit5 test. We try not to pollute the classpath, so
         # be very careful when adding new deps here.
+        #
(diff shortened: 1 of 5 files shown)


fmeum:
@shs96c Friendly ping :-)

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BUILD.bazel
```diff
@@ -58,11 +58,18 @@ java_library(
...
     ],
+    resource_strip_prefix = package_name(),
+    resources = ["META-INF/services/org.junit.jupiter.api.extension.Extension"],
     deps = [
         ":system-exit-toggle",
         # The only dependencies here are those required to run
         # a junit5 test. We try not to pollute the classpath, so
         # be very careful when adding new deps here.
+        #
+        # junit-jupiter-api is only a compilation dependency and not listed in JUNIT5_DEPS. This is
```
illicitonion, line 69: Should we add it to `JUNIT5_DEPS`?

illicitonion (approved):
LGTM, thanks!

java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BUILD.bazel
fmeum: Adding it could result in errors due to duplicates in `deps` if users declare it themselves. 

Luckily we don't need to add it as it is a runtime dependency of the Jupiter engine, which is always on the classpath.

Issue #211, opened 2023-10-01
Failed to find javaparser in runfiles


KrisJohnstone:
Attempting to convert a project and am getting a javaparser error.

**Expected Outcome:**
(At the very least a migrated listed of maven dependencies)

**Actual Outcome:**
11:03PM FTL external/contrib_rules_jvm/java/gazelle/configure.go:133 > could not start javaparser error="failed to start / connect to javaparser server: failed to find javaparser in runfiles"

**Files/Stepts to Reproduce:**
Workspace:
```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "io_bazel_rules_go",
    sha256 = "278b7ff5a826f3dc10f04feaf0b70d48b68748ccd512d7f98bf442077f043fe3",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_go/releases/download/v0.41.0/rules_go-v0.41.0.zip",
        "https://github.com/bazelbuild/rules_go/releases/download/v0.41.0/rules_go-v0.41.0.zip",
    ],
)

http_archive(
    name = "bazel_gazelle",
    sha256 = "d3fa66a39028e97d76f9e2db8f1b0c11c099e8e01bf363a923074784e451f809",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.33.0/bazel-gazelle-v0.33.0.tar.gz",
        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.33.0/bazel-gazelle-v0.33.0.tar.gz",
    ],
)

load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")

go_rules_dependencies()

go_register_toolchains(version = "1.20.5")

gazelle_dependencies()

http_archive(
    name = "contrib_rules_jvm",
    sha256 = "bd0f82def1879df85ff0a80767e6455911e1c9c1eac5db1de8f68dcccd4a3d7a",
    strip_prefix = "rules_jvm-0.18.0",
    url = "https://github.com/bazel-contrib/rules_jvm/releases/download/v0.18.0/rules_jvm-v0.18.0.tar.gz",
)

load("@contrib_rules_jvm//:repositories.bzl", "contrib_rules_jvm_deps")

[...]

PR #216, opened 2023-10-23
Update go dependencies

stevebarrau:
This should resolve https://github.com/bazel-contrib/rules_jvm/issues/211 for good.

The java dependencies changes are caused by a run of `./tools/update-dependencies.sh`.

PR #224, opened 2023-12-06
fix(java/gazelle): Use new gazelle runfiles API to locate javaparser

ar3s3ru:
This PR extends the work from @stevebarrau in #211 (still open at this time) to fix and simplify the new logic to locate the `javaparser` binary in runfiles, using the new Gazelle's `runfiles` package.

While this PR fixes the issue with locating `javaparser`, it produces a new error: `Cannot locate runfiles directory`.


stevebarrau:
IIUC the issue is not with java_stub_template.txt but with the gazelle runner shell wrapper @ar3s3ru (see https://github.com/bazelbuild/bazel-gazelle/pull/828). I inline your changes and got a green CI run in https://github.com/bazel-contrib/rules_jvm/pull/216/commits/711b68ceb76a38cf6c58a8998e4656114fee6d7d.

ar3s3ru:
Thanks for the ping @stevebarrau and linking the root issue.

I still see your PR is missing the handling of `JAVA_RUNFILES` environment variable. Only fixing the runfile location resolution won't make the plugin runnable (at least in my case)

Can we find a solution to get this merged?

Either your PR or mine, as long as you can include my commits so my work is part of the history, this took some effort 😅

ar3s3ru:
Thank you both @illicitonion @stevebarrau 🙌🏻 Glad to see the issue fixed 🎉

PR #225, opened 2023-12-12
set duplicate artifacts warning to none

ziad-fernride:


shs96c:
I think the better solution here is to make the `contrib_rules_jvm_deps` use the existing `lock_file` we have for this, and which we already use for the workspace-based builds.

shs96c:
The lock file is now being used in the `MODULE.bazel` file. Do we still need this PR?

shs96c:
Closing, as this has been awaiting an answer since 2021

PR #232, opened 2024-01-08
Update to stable download URL

illicitonion:


.bcr/source.template.json
```diff
@@ -1,5 +1,5 @@
 {
   "integrity": "",
   "strip_prefix": "{REPO}-{VERSION}",
-  "url": "https://github.com/{OWNER}/{REPO}/archive/refs/tags/{TAG}.tar.gz"
+  "url": "https://github.com/{OWNER}/{REPO}/releases/download/{VERSION}/${REPO}-${VERSION}.tar.gz"
```
fmeum, line 4: The current URL is https://github.com/bazel-contrib/rules_jvm/releases/download/v0.21.1/rules_jvm-v0.21.1.tar.gz, so I think you need to replace `VERSION` with `TAG`.

illicitonion: Gotta love that the v-adding/stripping behaviour. Good catch, thanks!

fmeum (approved):

PR #233, opened 2024-01-09
Add plugin support to Spotbugs

mwindmark:
Adds support for using plugins with SpotBugs, see [documentation for common plugins](https://spotbugs.readthedocs.io/en/stable/links.html#spotbugs-plugins). 

This can for example be used to activate the [Find Security Bugs](https://find-sec-bugs.github.io/) plugin.

```
spotbugs_config(
    name = "spotbugs-custom-config",
    effort = "default",
    fail_on_warning = True,
    plugin_list = [
        "@maven//:com_h3xstream_findsecbugs_findsecbugs_plugin",
    ],
    visibility = ["//visibility:public"],
)
```


mwindmark:
> This looks good, thanks for the contribution! Would you be able to add an example or test which uses this new functionality?

Thanks! I saw that you recently added the example `tests_and_lints` so I extended that with Spotbugs. See the changes in [6c6a3cc](https://github.com/bazel-contrib/rules_jvm/pull/233/commits/6c6a3cc4a18e741d1923f77085915e0219c5fbc6)

illicitonion (approved):
Perfect, thanks so much!

PR #238, opened 2024-01-11
Rename format pipeline

illicitonion:
Our CI checks aren't currently required to merge.

As far as I can tell, this name key is the only thing you can use to describe whether a check is required, and the fact that it matches the name for the other GitHub action means we can't make both required.


shs96c (approved):
Let's see if that works

PR #248, opened 2024-02-16
ClasspathParser: handle null packageName in fullyQualify

gaurav-narula:


shs96c (approved):
LGTM. Thank you for catching this, and for the test!

PR #250, opened 2024-02-26
update rules_jvm_external to 6.0

lyao-77:
We use `java_export` in our project and need to use some new feature that was added in to newly released `rules_jvm_external`

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -51,7 +51,7 @@ bazel_dep(
 )
 bazel_dep(
     name = "rules_jvm_external",
-    version = "5.3",
+    version = "6.0",
 )
 bazel_dep(
     name = "rules_proto",


MODULE.bazel
```diff
@@ -51,7 +51,7 @@ bazel_dep(
 )
 bazel_dep(
     name = "rules_jvm_external",
-    version = "5.3",
+    version = "6.0",
```
shs96c, line 54: Could you please update the `repositories.bzl` file too? The versions of `rules_jvm_external` needs to match both here and there.

shs96c (changes requested):
Thank you for the PR!

I'll investigate the CI failure, as I think it's probably a bug in `rules_jvm_external` rather than anything to do with this PR. You don't need to worry about that error.

shs96c:
Another PR (#288) updated `contrib_rules_jvm` to `rules_jvm_external` 6.2. Closing this, but thank you for your PR

PR #254, opened 2024-04-09
gazelle: Fail on unknown import

stevebarrau:
If a Java import cannot be resolved, the extension should fail.


illicitonion (approved):
Thanks!

PR #258, opened 2024-05-01
Set bazel version when testing bcr support

illicitonion:
Looks like this became required since the last release.


shs96c (approved):
LGTM

PR #259, opened 2024-05-01
Fix "failed to call RLocation: runfile (...)/Main: file does not exist" 

bartoszpop:
Fix "bazel run gazelle" failing on Windows with the following error:
> 1:57PM FTL external/contrib_rules_jvm/java/gazelle/configure.go:139 > could not start javaparser error="failed to start / connect to javaparser server: failed to find javaparser in runfiles: failed to call RLocation: runfile contrib_rules_jvm/java/src/com/github/bazel_contrib/contrib_rules_jvm/javaparser/generators/Main: file does not exist" (#257)


illicitonion (approved):
Thanks!

Issue #260, opened 2024-05-01
Gazelle fails on Windows with "panic: runtime error: invalid memory address or nil pointer dereference"


bartoszpop:
`bazel run gazelle` fails on Windows with the following error:
> INFO: Running command line: bazel-bin/external/bp_bazel/tools/gazelle/gazelle.exe
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x58 pc=0x9efb76]
goroutine 1 [running]:
github.com/bazel-contrib/rules_jvm/java/gazelle/javaconfig.(*Config).NewChild(...)
        external/contrib_rules_jvm/java/gazelle/javaconfig/config.go:59
github.com/bazel-contrib/rules_jvm/java/gazelle.(*Configurer).Configure(0xc0002c0750, 0xc000221080, {0xc0004e9ad0, 0x29}, 0x0)
        external/contrib_rules_jvm/java/gazelle/configure.go:84 +0x316
github.com/bazelbuild/bazel-gazelle/walk.configure({0xc0002adcc0, 0xa, 0x29?}, 0xc000056500?, 0xc000220fc0?, {0xc0004e9ad0, 0x29}, 0x0)
        external/bazel_gazelle/walk/walk.go:289 +0x383
github.com/bazelbuild/bazel-gazelle/walk.Walk.func1(0xc000220fc0, {0xc000056500, 0x55}, {0xc0004e9ad0, 0x29}, 0x1)
        external/bazel_gazelle/walk/walk.go:145 +0x2d6
github.com/bazelbuild/bazel-gazelle/walk.Walk.func1(0xc000220f00, {0xc0001bb600, 0x4e}, {0xc0004e9230, 0x22}, 0x1)
        external/bazel_gazelle/walk/walk.go:169 +0x8b7
github.com/bazelbuild/bazel-gazelle/walk.Walk.func1(0xc000220300, {0xc0003e5f80, 0x39}, {0xc0004f4c10, 0xd}, 0x1)
        external/bazel_gazelle/walk/walk.go:169 +0x8b7
github.com/bazelbuild/bazel-gazelle/walk.Walk.func1(0xc000220300, {0xc0001ae8d0, 0x2b}, {0x0, 0x0}, 0x0)
        external/bazel_gazelle/walk/walk.go:169 +0x8b7
[...]

PR #261, opened 2024-05-01
Fix "panic: runtime error: invalid memory address or nil pointer dereference"

bartoszpop:
Fix `bazel run gazelle` failing on Windows with
> panic: runtime error: invalid memory address or nil pointer dereference

because of the incorrect path separator when accessing the Configs map (#260).

### DIFF
--- a/java/gazelle/configure.go
+++ b/java/gazelle/configure.go
@@ -3,8 +3,8 @@ package gazelle
 import (
 	"flag"
 	"fmt"
-	"strings"
 	"path/filepath"
+	"strings"
 
 	"github.com/bazel-contrib/rules_jvm/java/gazelle/javaconfig"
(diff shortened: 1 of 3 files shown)


illicitonion:
Thanks for your fix! (And all your fixes - I've been working today on getting our CI running on Windows so we can catch this kind of problem in the future and make sure things work and keep working).

I think https://github.com/bazel-contrib/rules_jvm/pull/264 is a slightly more targeted fix - does that look good to you?

bartoszpop:
> Thanks for your fix! (And all your fixes - I've been working today on getting our CI running on Windows so we can catch this kind of problem in the future and make sure things work and keep working).
> 
> I think #264 is a slightly more targeted fix - does that look good to you?

@illicitonion LGTM, I checked on my local and your fix also works.

Issue #262, opened 2024-05-01
Gazelle fails on Windows with "java.nio.file.InvalidPathException: Illegal char <:> at index 2"


bartoszpop:
`bazel run gazelle` fails on Windows with the following error:
> INFO GrpcServer - Set of classes in c:\users\bpopiela\_bazel~1\q6mbxjwl\execroot\my-pro~1\bazel-~1\x64_wi~2\bin\external\bp_bazel\tools\gazelle\gazell~2.run\my-pro~1\src\test\java has no package
ERROR ClasspathParser - JavaTools failed to parse [SimpleFileObject[C:\Users\bpopiela\IdeaProjects\my-project\src\test\java\SomeClassTest.java]], skipping file
java.nio.file.InvalidPathException: Illegal char <:> at index 2: /C:/Users/bpopiela/IdeaProjects/my-project/src/test/java/SomeClassTest.java
        at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
        at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
        at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
        at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
        at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:232)
        at java.base/java.nio.file.Path.of(Path.java:147)
        at java.base/java.nio.file.Paths.get(Paths.java:69)
        at com.github.bazel_contrib.contrib_rules_jvm.javaparser.generators.ClasspathParser$ClassScanner.visitCompilationUnit(ClasspathParser.java:183)
        at com.github.bazel_contrib.contrib_rules_jvm.javaparser.generators.ClasspathParser$ClassScanner.visitCompilationUnit(ClasspathParser.java:171)
        at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCCompilationUnit.accept(JCTree.java:614)
        at com.github.bazel_contrib.contrib_rules_jvm.javaparser.generators.ClasspathParser.parseFileGatherDependencies(ClasspathParser.java:161)
        at com.github.bazel_contrib.contrib_rules_jvm.javaparser.generators.ClasspathParser.parseClasses(ClasspathParser.java:148)
[...]

PR #263, opened 2024-05-01
Fix "java.nio.file.InvalidPathException: Illegal char <:> at index 2"

bartoszpop:
Fix `bazel run gazelle` failing with `java.nio.file.InvalidPathException: Illegal char <:> at index 2` (#262).


illicitonion (approved):
Thank you!

PR #266, opened 2024-05-02
Shard CI

illicitonion:
CI is currently unbearably slow because it runs each config sequentially.

Instead, run them in parallel. This also opens up matrix support to be able to run more OSes (because we should test on Windows, which increasingly is starting to work :))


shs96c (approved):
CI seemed to pass. LGTM on that basis

PR #267, opened 2024-05-02
Only call rlocation once

illicitonion:
This leads to slightly faster runtime, and slightly less confusing error messages if the file can't be found.


shs96c (approved):
LGTM

PR #270, opened 2024-05-04
Use an OS-agnostic path separator for the srcs attribute

bartoszpop:
This pull request updates the srcs attribute with "/" regardless of the OS not to update the BUILD file when Gazelle is run on a different OS and to correctly trim the parent path (#269).

### DIFF
--- a/java/gazelle/generate.go
+++ b/java/gazelle/generate.go
@@ -257,7 +257,7 @@ func (l javaLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 			srcs := make([]string, 0, allTestRelatedSrcs.Len())
 			for _, src := range allTestRelatedSrcs.SortedSlice() {
 				if _, ok := separateTestJavaFiles[src]; !ok {
-					srcs = append(srcs, strings.TrimPrefix(src.pathRelativeToBazelWorkspaceRoot, args.Rel+"/"))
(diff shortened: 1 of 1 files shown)


java/gazelle/generate.go
```diff
@@ -257,7 +257,7 @@ func (l javaLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 			srcs := make([]string, 0, allTestRelatedSrcs.Len())
 			for _, src := range allTestRelatedSrcs.SortedSlice() {
 				if _, ok := separateTestJavaFiles[src]; !ok {
-					srcs = append(srcs, strings.TrimPrefix(src.pathRelativeToBazelWorkspaceRoot, args.Rel+"/"))
+					srcs = append(srcs, strings.TrimPrefix(filepath.ToSlash(src.pathRelativeToBazelWorkspaceRoot), filepath.ToSlash(args.Rel+"/")))
```
illicitonion, line 260: I believe `args.Rel` is always forward-slashed?

```suggestion
					srcs = append(srcs, strings.TrimPrefix(filepath.ToSlash(src.pathRelativeToBazelWorkspaceRoot), args.Rel+"/"))
```

illicitonion (approved):
Thanks for the fix!

LGTM with one suggestion

PR #272, opened 2024-05-07
Replace warnings with empty files

illicitonion:
Right now lots of our tests assert that there's a warning about a missing file.

Instead, just provide a valid dummy file.


java/gazelle/testdata/annotations/expectedStderr.txt
```diff
@@ -1 +0,0 @@
-{"level":"warn","_c":"maven-resolver","error":"open %WORKSPACEPATH%/maven_install.json: no such file or directory","message":"not loading maven dependencies"}
```
gibfahn, line 1: Is it worth leaving one test that checks for this error message?

illicitonion: It is, but for cross-platform reasons this is annoyingly fiddly. If it's ok with you, I'd like to merge this as-is, then add a feature to gazelle itself which would let us write this test reasonably, then add a new test which does this assertion?

gibfahn: Ahh, so this is also part of the "Windows fixes" work right?

gibfahn (approved):

PR #276, opened 2024-05-08
gazelle: Populate plugins attributes with annotation processors

illicitonion:


stevebarrau (approved):
Thanks @illicitonion. I see this being useful for things like auto (in your example), Lombok and many other compilation plugins.

PR #282, opened 2024-06-18
Adding a plugin adds it to plugins

illicitonion:
Previously if there was an existing plugins attribute, and an additional plugin was added, the old value would get preserved rather than merged.


gibfahn (approved):

shs96c:
@illicitonion, could you please rebase on `main` so we can get the failing build working?

PR #284, opened 2024-06-24
Escape invalid XML chars in testCase and testSuite names

sfc-gh-abalik:
Fix for https://github.com/bazel-contrib/rules_jvm/issues/285

With a [ParameterizedTest](https://junit.org/junit5/docs/5.0.2/api/org/junit/jupiter/params/ParameterizedTest.html) it is possible to generate test case names with invalid XML characters. E.g.

```
@ParameterizedTest
@ValueSource(strings = {"Weird\bname"})
void testFoo(String input) {
    ...
}
```

The test case name is derived from the input string, which in this case contains an invalid XML char `\b`. When tools (like IDEs) try to parse the junit XML, they'll fail due to the invalid char.

Fix it by escaping test case names. It looks like [test suite names](https://junit.org/junit5/docs/5.0.3/api/org/junit/jupiter/api/DisplayName.html) could also contain invalid XML chars, so I'm escaping those as well.

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestCaseXmlRenderer.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestCaseXmlRenderer.java
@@ -47,7 +47,7 @@ class TestCaseXmlRenderer {
     }
 
     xml.writeStartElement("testcase");
-    xml.writeAttribute("name", name);
+    xml.writeAttribute("name", escapeIllegalCharacters(name));
     xml.writeAttribute("classname", LegacyReportingUtils.getClassName(testPlan, id));
 
     /* @Nullable */ Duration maybeDuration = test.getDuration();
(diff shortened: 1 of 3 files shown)


illicitonion (approved):
Thanks!

illicitonion:
The CI failure is spurious, but is blocked on https://github.com/bazelbuild/rules_jvm_external/pull/1173 getting released - hopefully it will get released soon!

sfc-gh-abalik:
Looks like https://github.com/bazelbuild/rules_jvm_external/pull/1173 got released.
I think we just need to rebase after https://github.com/bazel-contrib/rules_jvm/pull/288 is merged.

shs96c:
@sfc-gh-abalik, with #288 we now have the CI builds working properly again. Could you please rebase on top of current `HEAD` of `main` so that we can land this change? Thank you for your patience as we were sorting that out!

shs96c (approved):
LGTM. Thank you!

PR #289, opened 2024-07-11
refactor: remove runner attribute from create_jvm_test_suite

francine-blanc:
In `create_jvm_test_suite.bzl`, the `runner` attribute is not utilised within the `define_test` function. By the time we reach the `create_jvm_test_suite` macro, this attribute is redundant because the specific test runner can be directly specified within the `define_test` function provided by the user. So, this PR removes references to the runner from the macro.

### DIFF
--- a/java/private/create_jvm_test_suite.bzl
+++ b/java/private/create_jvm_test_suite.bzl
@@ -9,13 +9,6 @@ def _is_test(src, test_suffixes, test_suffixes_excludes):
             return True
     return False
 
-# If you modify this list, please also update the `_TEST_GENERATORS`
-# map in `java_test_suite.bzl`.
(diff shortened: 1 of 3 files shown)


shs96c (approved):
Thank you! LGTM

sitaktif:
A consequence of this changes is that for existing users, the `runner` attribute is now going to be passed to the `define_test` function (via `kwargs`). Just checking that this change is intended and that it's acceptable in terms of backward incompatibility.

shs96c:
That’s fine. This macro is seldom used, as it’s a building block for the
higher level suite macros.

francine-blanc:
@shs96c would you be able to trigger the pipeline again please?

PR #290, opened 2024-07-12
Update deps freezing script to work with bzlmod

shs96c:


tools/freeze-deps.py
```diff
@@ -73,18 +81,33 @@
...
-    output.writestr(zinfo, "\n".join(lines))
-
+# Generate a compat.bzl file. This is only needed for workspace-based builds, but the
+# file is not generated in a `bzlmod`-created `@maven`.
+lock_file = root / "imported_maven_install.json"
+
+compat_content = """load("@contrib_rules_jvm_deps//:compat_repository.bzl", "compat_repository")
+
+def compat_repositories():
+    """
```
mkosiba, line 91: do you want to add a `pass` here for the super-weird case of there not being any artifacts?

mkosiba (approved):

shs96c:
If we’re freezing deps we must have artifacts.

tools/freeze-deps.py
shs96c: I don't think we'd be freezing deps if we have no artifacts….

PR #294, opened 2024-08-07
Add timestamp and hostname to testsuite xml

### LINKED ISSUE #293: Junit5: XML output does not adhere to JUnit schema
Taking a look at TestSuiteXmlRenderer, there are some fields marked required by the JUnit schema that are missing, for example timestamp which by the schema: https://windyroad.com.au/dl/Open%20Source/JUnit.xsd is a required attribute. 

```
<xs:attribute name="timestamp" type="ISO8601_DATETIME_PATTERN" use="required">
<xs:annotation>
<xs:documentation xml:lang="en">when the test was executed. Timezone may not be specified.</xs:documentation>
</xs:annotation>
</xs:attribute>
```

brian-mcnamara:
Fixed #293 

Per https://windyroad.com.au/dl/Open%20Source/JUnit.xsd timestamp and hostname are required attributes on the testsuite tag.

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestSuiteXmlRenderer.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestSuiteXmlRenderer.java
@@ -3,6 +3,8 @@ package com.github.bazel_contrib.contrib_rules_jvm.junit5;
 import static com.github.bazel_contrib.contrib_rules_jvm.junit5.SafeXml.escapeIllegalCharacters;
 import static com.github.bazel_contrib.contrib_rules_jvm.junit5.SafeXml.writeTextElement;
 
+import java.net.InetAddress;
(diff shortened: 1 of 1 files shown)


java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestSuiteXmlRenderer.java
```diff
@@ -67,4 +71,12 @@ public void toXml(XMLStreamWriter xml, TestData suite, Collection<TestData> test
 
     xml.writeEndElement();
   }
+
+  private String getHostname() {
```
brian-mcnamara, line 75: The spec does callout that localhost can be used, I do not have a need to know the host, but figure I would attempt to set it in case anyone does need it

shs96c (approved):
LGTM. I'm slightly worried about the potential network call during the hostname lookup, but the exception should handle that case. 

Thank you for the PR!

PR #297, opened 2024-08-29
Gazelle: detect any directory with /test in path as test package

kusaeva:
We have not very java-like structure in our monorepo:
Screen Shot 2024-08-29 at 16 53 15
So current func IsTestPackage doesn't detect our tests as tests and we have `found MULTIPLE results in rule index` error because of the same package in test classes.
We propose the fix, so files in packages like `//common/test` will also fit as test packages.

### DIFF
--- a/java/gazelle/private/java/java.go
+++ b/java/gazelle/private/java/java.go
@@ -21,7 +21,7 @@ func IsTestPackage(pkg string) bool {
 		}
 	}
 
-	return strings.Contains(pkg, "/test")
+	return strings.Contains(pkg, "/test/") || strings.HasSuffix(pkg, "/test")
 }
 
 // This list was derived from a script along the lines of:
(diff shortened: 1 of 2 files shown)


java/gazelle/private/java/java.go
```diff
@@ -21,7 +21,7 @@ func IsTestPackage(pkg string) bool {
 		}
 	}
 
-	return strings.Contains(pkg, "/test/")
+	return strings.Contains(pkg, "/test")
```
illicitonion, line 24: This would cause issues for packages with names like `testutils`, but probably a trailing `test` package segment makes sense to check for - WDYT?

```suggestion
	return strings.Contains(pkg, "/test/") || strings.HasSuffix(pkg, "/test")
```

illicitonion:
Thanks for the contribution!

java/gazelle/private/java/java.go
kusaeva: You're totally right! Thanks!

illicitonion (commented):
Please could you add a couple of new test-cases to https://github.com/bazel-contrib/rules_jvm/blob/main/java/gazelle/private/java/java_test.go (one for `test` at the end, one for `testutil` in the middle) to make sure this behaviour doesn't regress? Otherwise, happy to merge! Thanks!

illicitonion (approved):
Thanks again!

PR #298, opened 2024-10-03
Adds several attributes to spotbugs rule

sfc-gh-mgalindo:


java/private/spotbugs.bzl
```diff
@@ -94,6 +113,14 @@ spotbugs_test = rule(
                 SpotBugsInfo,
             ],
         ),
+        # NOTE: this will always be just one file, but we use a label_list instead of a label
+        # so that we can pass a glob. In cases where no baseline file exist, the glob will simply
+        # be empty. Using a label instead of a label_list would force us to create a baseline file
+        # for all targets, even if theres no need for one.
+        "baseline_file": attr.label_list(
```
shs96c, line 120: Please just make this an `attr.label`, and you don't need to set the `default` since it will automatically be `None`. The check that's `if len(baseline_files) > 0:` will then just be `if ctx.attr.baseline_file:` which will be fine.

The logic of this statement is also incorrect if `--incompatible_disallow_empty_glob` is set, so I'm not sure this is really helping.

shs96c (changes requested):
Thank you for your patience waiting for this review. There's one small change I've asked for, and then I'd be happy to merge this into the tree.

PR #303, opened 2024-10-22
Optimize junit dependencies for each test target

brian-mcnamara:
This change simplifies the dependency list for the junit suite test targets by creating a export-only library used to group the dependencies together via an export-only java_library. The primary reason for this is to simplify the project model for IntelliJ using the new Bazel plugin. With this change, the IntelliJ model does not duplicate the dependencies for each test class in the Aspects and the internal project model and instead only refers to the java library.

### DIFF
--- a/java/private/create_jvm_test_suite.bzl
+++ b/java/private/create_jvm_test_suite.bzl
@@ -91,6 +91,26 @@ def create_jvm_test_suite(
 
     tests = []
 
+    # Optimization for classpath, reduces the duplicate dependencies for each test to instead rely on one target
+    deps_lib_name = "%s-test-deps-lib" % name
+    define_library(
+        name = deps_lib_name,
+        exports = deps,
+        visibility = ["//visibility:private"],
+        tags = tags,
+        testonly = True,
+        **library_attrs
+    )
+    runtime_deps_lib_name = "%s-test-runtime-deps-lib" % name
+    define_library(
+        name = runtime_deps_lib_name,
+        exports = runtime_deps,
+        visibility = ["//visibility:private"],
+        tags = tags,
+        testonly = True,
+        **library_attrs
+    )
+
     for src in test_srcs:
         suffix = src.rfind(".")
         test_name = src[:suffix]
@@ -101,9 +121,9 @@ def create_jvm_test_suite(
             size = size,
             srcs = [src],
             test_class = test_class,
-            deps = deps,
+            deps = [":" + deps_lib_name],
             tags = tags,
-            runtime_deps = runtime_deps,
+            runtime_deps = [":" + runtime_deps_lib_name],
             visibility = ["//visibility:private"],
             **kwargs
         )


brian-mcnamara:
Thanks @shs96c for taking a look! Is there plans to publish a release in the near future?

brian-mcnamara:
@shs96c would a release be possible, or is there one planned in the near future?

shs96c:
We can push a release, but I'd like us to fix the artifact hashing algorithms first. My day job has been pretty busy, so I've not had much time to focus on that.

brian-mcnamara:
Totally understandable. Ill update us to use a commit in the mean time. Thanks for the update!

java/private/create_jvm_test_suite.bzl
```diff
@@ -91,6 +91,26 @@ def create_jvm_test_suite(
 
     tests = []
 
+    # Optimization for classpath, reduces the duplicate dependencies for each test to instead rely on one target
+    deps_lib_name = "%s-test-deps-lib" % name
+    define_library(
```
honnix, line 96: Not entirely sure but will these two new jars include test resources? We are seeing an issue when using liquibase in a test, as it complains about duplicated resources files from different jars.

brian-mcnamara: It will not create any jar as there are no sources to this library. It should only be a bazel construct which allows aspects to simplify the class path resolution.

honnix: Hmm, that's not what we observe. Yes it doesn't container any Java source code, but it contains test resources. Could it be https://github.com/brian-mcnamara/rules_jvm/blob/d899e276ff1a4a9b98272ed6cd31991f09d7839a/java/private/create_jvm_test_suite.bzl#L16?

honnix: Does something like https://github.com/bazel-contrib/rules_jvm/compare/main...honnix:rules_jvm:filter-out-resources?expand=1 make sense to you?

brian-mcnamara: Taking a look, want to see if data is actually needed at all, or whether we could just remove this all together.

honnix: Please feel free to ignore my draft and go ahead fixing it. Thank you for looking into this issue. 🙏

brian-mcnamara: Sorry for the delay, my environment was being slow yesterday, still looking at it :)

brian-mcnamara: @honnix , I was creating a test case and seems like this was the case even before my change: https://github.com/bazel-contrib/rules_jvm/compare/main...brian-mcnamara:rules_jvm:resourceRemoval?expand=1
Both the ["test-lib"](https://github.com/brian-mcnamara/rules_jvm/blob/main/java/private/create_jvm_test_suite.bzl#L80) and actuall [test](https://github.com/brian-mcnamara/rules_jvm/blob/main/java/private/create_jvm_test_suite.bzl#L99) have the resource as part of it. 

Is it the case that your test suites only contain test sources (such that you never get into the logic here: https://github.com/brian-mcnamara/rules_jvm/blob/main/java/private/create_jvm_test_suite.bzl#L71

honnix: Sorry I'm out of town these days and don't have computer with me. I can look into it in depth tomorrow.

> Is it the case that your test suites only contain test sources (such that you never get into the logic here:

I don't think that is the case.

honnix: That said, I think it is still not ideal having things duplicated. Maybe resources should not be included in `-test-lib` jar (I don't know the reason why they are explicitly included).

honnix: Looks like it was like this since the beginning https://github.com/bazel-contrib/rules_jvm/commit/7c008f2e785812ac2bb2898aa47e8082870d39de#diff-7e53ad4b884087fec5fdb0e9b2cf5d2f4fd2390dc68161d0a09878977a34414b

brian-mcnamara: Ooph, not great, but thanks for confirming my test was not doing something crazy. I agree it does not make sense to have resources included multiple times. Let me at least propose a PR and see where it goes :)

brian-mcnamara: https://github.com/bazel-contrib/rules_jvm/pull/338 created PR, will see what the maintainers think of the approach.

PR #307, opened 2024-11-04
Bump lowest supported Java version to 8 in our build files

shs96c:
Doing this prevents warnings about Java 8 being no longer supported when using a modern JDK and compiling our own supporting code.


java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BUILD.bazel
```diff
@@ -52,11 +52,11 @@ java_library(
         ],
     ),
     javacopts = [
-        # Target Java 8, no matter which JDK we're on so that we
+        # Target Java 11, no matter which JDK we're on so that we
```
shs96c, line 55: This is fine, since the SecurityManager we want to use is in Java 11

stevebarrau (approved):

PR #312, opened 2024-11-22
allow passing empty 'package' to 'java_test_suite'

mattnworb:
If for whatever reason one has test classes in the unnamed package, the current logic for detecting the full classname will return something like `src.test.java.FoobarTest` when you specify `package = ""` because of the truthiness test in `get_class_name`.


shs96c (approved):
Thank you for your patience waiting for this review. LGTM, and thank you for the PR!

Issue #126, opened 2023-01-30
FR: Gazelle Kotlin Support


rob-pomelo:
Hi is there any thought / interest in adding kotlin support to the gazelle extension?

Have briefly started to look at the changes required and curious if its on others radar.

Thanks!

illicitonion:
Would gladly review some PRs (with tests) to enable this!

MorganR:
I've put up a (very messy, don't use it) PR up (#313 )  for discussion on this

MorganR:
I'd also be curious what @alexeagle thinks about merging Kotlin support in with the Java plugin, given his experience from the independent Kotlin plugin.

restingbull:
Hm. We are generally trying to keep our maintenance footprint small, and adding Go to the dependencies is a bit non-ideal for us.

@Bencodes @nkoroste-- Thoughts?

MorganR:
@shs96c, I understand why it seems like rules_kotlin would be a good spot for this, but I actually think that if someone is using Kotlin specifically with the JVM, then the best place for handling those rules is in the same plugin that handles the Java rules. I put more thoughts about why this is the case in the draft PR I linked above.

The only new dependency it adds to rules_jvm is one new Maven dependency on the "kotlin" package for parsing the Kotlin files.

shs96c:
I am quiet persuaded by the argument that kotlin and java source is often co-mingled in a way that other JVM-based languages don't seem to do.

MorganR:
I'm aware of the existing extension, but had trouble using it in a repo with mixed Kotlin and Java. 

@alexeagle or @jbedard, do either of you think there is merit to merging the two? I discussed this a bit in the [related issue](https://github.com/bazel-contrib/rules_jvm/pull/313), but would especially like your insights having worked on the Kotlin plugin.

MorganR:
Also I agree with you Alex that having Gazelle interact with subprocesses directly would be a nice improvement, though it doesn't seem required. The Java extension currently gets around this by starting its own Java-based gRPC server for parsing.

However, that makes me wonder if you could extend that idea further to something akin to an LSP or [BSP](https://build-server-protocol.github.io/), but for Gazelle? For larger codebases, it might be helpful to have a long-running process that can handle incremental changes, rather than processing everything from scratch on each invocation. I'm not sure if Gazelle already does some kind of local file caching or other technique to improve this.

alexeagle:
[...]

PR #313, opened 2024-12-01
DRAFT: Very rough in-progress work on Kotlin support in the Gazelle plugin

MorganR:
Do not merge this! It's for discussion. This has soo many caveats right now. It is ugly, is messy, is incomplete, has debug statements everywhere, has commented out code, etc. 

I'm putting this up anyway so that people can take an early look at the overall approach, and give any general comments or guidance.

The overall approach:

* Add a new Gazelle directive `jvm_kotlin_enabled`, which defaults to `false`, to optionally enable Kotlin processing. Use this flag to optionally look for Kotlin files in addition to Java files within packages. If a Kotlin file is present in a package, then we use the `kt_jvm_library` rule instead of `java_library` rule. Most remaining Go logic is unchanged.
* Add a `KtParser` to the `JavaParser` gRPC server. This uses the Kotlin compiler's parser to process Kotlin files, and extract the same data that is extracted by the Java-oriented `ClasspathParser`.
  * This file currently handles detecting a main function, imported types and packages, and class declarations. It has partial coverage for exported types, and has a bunch of commented out code where I'm trying to figure out the best way to tackle detecting fully-qualified types and packages throughout the file. I also haven't yet added support for any annotation handling.

# Questions for reviewers

1. Overall, WDYT? If people think this is worth adding, I'll make a series of smaller, cleaner PRs to actually get this done.
2. The plugin, its directives, packages, and classes often use the name `java`. By adding Kotlin, I'm somewhat implicitly changing the interpretation of this to something more like `jvm`. Any thoughts yay or nay on name changes? That's not an all or nothing thing. For example, maybe we rename the Gazelle `Language` to `jvm`, but keep the existing directive names to avoid breaking users. Or, we add new `jvm` directives, but continue to support the old `java` ones as well. Lots of other possibilities too.

# FAQ

**Why do this here when there is [another Kotlin plugin](https://github.com/aspect-build/aspect-cli/tree/main/gazelle/kotlin)?**

A few reasons:

1. If you're using Kotlin + Java, and exclusively using Kotlin with the JVM, then I'm not sure that it actually makes sense to treat Kotlin as a separate language from Gazelle's perspective. That makes sense for two languages like protobuf and Java, which have a strictly one-way dependency graph. However, in my experience, Java and Kotlin code frequently depend on each other, sometimes from separate packages, and sometimes within the same package. If dependencies are only _between_ packages, then I can see separate Gazelle plugins potentially working with Gazelle's [cross-resolve feature](https://github.com/bazel-contrib/bazel-gazelle/issues/771), though this seems more intended for entirely separate resolution methods, not inter-dependent language plugins. Additionally, I don't see a good way to handle intra-package Java & Kotlin dependencies. If the Java plugin knows nothing about Kotlin, it will try to produce a java_library rule for the Java files, which will lack some Kotlin-specific information. A Kotlin plugin would then need to replace this rule with a Kotlin one, and extend it to handle the extra information from the Kotlin files. This seems a bit hacky and error prone. Conversely, adding Kotlin support directly to the Java plugin is turning out to be relatively painless (minus me trying to learn how to use the Kotlin parser).
3. There is an open [support request for it](https://github.com/bazel-contrib/rules_jvm/issues/126).

**Why add the `KtParser` to the same gRPC request handler, instead of a separate one for Kotlin?**

I'd be happy to change this if others feel strongly, but we're ultimately trying to produce the exact same format of output data for Java and Kotlin, and merge them into a single set of data per package. We could either make two separate gRPC calls from Go land and then merge the data there, or make a single call and merge the data in Java land. That seemed like less boilerplate (defining new RPC requests) and less overhead (fewer network calls, even if they're to localhost). 

**Why add the `KtParser` to the same Java parsing server, instead of a separate server entirely?**

See above, but also this would be even more resource overhead. No need to kick off multiple JVM processes if we can do it all in one.


MorganR:
I should note: I also haven't yet added support for tests

shs96c:
I think this has been in limbo long enough. Let's push ahead with this, and we'll see what happens :) That okay with you @MorganR?

MorganR:
I would definitely want to tidy some in progress pieces before merging it, but happy to move forwards if you like the overall approach!

I can start sending some small & tidy PRs over if that works for you?

alexeagle:
@MorganR we're strongly considering donating the implementation from aspect-cli to this repo. Would you like to discuss joining efforts? Maybe on Bazel slack?

MorganR:
Hey folks, I apologies but I had to set this down for a while. I can get back into it now. @alexeagle, I've messaged you on Slack.

shs96c:
@MorganR, just another thank you for this! Your initial work landed in #375 and we've been iterating since then!

MorganR:
Aw that's great to hear it was useful in the end! Thank you for taking up the mantle 🙏 I know it had quite a ways to go when I last touched it, so major kudos to you for getting it over the line.

PR #315, opened 2024-12-02
Only run lint_tests on java_library not java_export

### LINKED ISSUE #314: java_export creates spotbugs/checkstyle rule for dependencies
https://github.com/bazel-contrib/rules_jvm/blob/50fec7933b7e5136a2c412b85990d024fce19698/java/private/library.bzl#L75

Runs the checkstyle/spotbugs on all sources in java_export which therefore includes all non-`java_export` targets.
This can be problematic if you include something like protobuf which will clearly violate spotbugs.

The fix is to have the checkstyle/spotbugs operate on the inner `java_library` crated by java_export

fzakaria:
Running list_tests on java_export will run afoul by trying to run spotbugs on included dependency that will clearly fail like `java_proto_library`.

Instead, the lint tests should only operate on the sources that comprise the `java_library` the export target eventually creates.

fixes #314

### DIFF
--- a/java/private/library.bzl
+++ b/java/private/library.bzl
@@ -71,7 +71,12 @@ def java_test(name, **kwargs):
 
 def java_export(name, maven_coordinates, pom_template = None, deploy_env = None, visibility = None, **kwargs):
     """Adds linting tests to `rules_jvm_external`'s `java_export`"""
-    create_lint_tests(name, **kwargs)
+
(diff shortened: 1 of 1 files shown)


fzakaria:
Validated this locally with `--override_module`

shs96c (approved):
LGTM. Thank you for the PR!

Issue #327, opened 2025-03-10
BazelJUnitOutputListener Logs Empty TestSuite on Test Suite Level Failure


amishra-u:
### Issue
When a test suite fails at the suite level, such as when a @BeforeClass method fails in a test class, BazelJUnitOutputListener logs an empty <testsuite> instead of properly capturing the failure. While Bazel logs the output for each test in the test class with the test result reflecting the test suite failure.

### Steps to Reproduce:
- Create a JUnit test class with a @BeforeClass method that fails (e.g., throws an exception).
```
  public static class TestSuiteError {
    @BeforeClass
    public static void setup() {
        throw new RuntimeException("test suite error");
    }

    @Test
    public void test1() {
      System.out.println("test1");
    }

    @Test
    public void test2() {
      System.out.println("test2");
    }
  }
```
- Check the XML output generated by BazelJUnitOutputListener. The <testsuite> corresponding to this class will be empty.

### Expected Behavior:
- Each test in the test suite should be logged with same error log as parent test suite. Idenitical to Bazel XmlWriter.

### Actual Behavior:
- BazelJUnitOutputListener logs an empty <testsuite>.

PR #300, opened 2024-10-10
Add support for @AfterAll in XML report

fzakaria:
Writing a simple test like
```java
package com.library;

import org.junit.jupiter.api.AfterAll;
import org.junit.jupiter.api.Test;

public class AlwaysFailTest {
    @Test
    public void doNothingTest() {
        System.out.println("Hi there!");
    }

    @AfterAll
    public static void alwaysFail() {
        throw new RuntimeException("Always failing.");
    }
}
```

Produces an XML that seems to look like it succeeds.
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="com.library.AlwaysFailTest" timestamp="2024-10-10T21:19:11.522176Z" hostname="KHK9NLVQGN" tests="2" failures="0" errors="0" disabled="0" skipped="0" package="">
    <properties/>
    <testcase name="doNothingTest" classname="com.library.AlwaysFailTest" time="0.03">
      <system-out><![CDATA[Hi there!
]]></system-out>
    </testcase>
  </testsuite>
</testsuites>
```

Bazel itself correctly identifies it fails. The Junit4 reporter also included with Bazel natively correclty reports the failure in the XML output.

Augment the Junit listener to add support for static methods and add them to the JUnit output.

Doing so produces the following XML.
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="com.library.AlwaysFailTest" timestamp="2024-10-10T21:49:02.096648Z" hostname="KHK9NLVQGN" tests="3" failures="1" errors="0" disabled="0" skipped="0" package="">
    <properties/>
    <testcase name="com.library.AlwaysFailTest" classname="com.library.AlwaysFailTest" time="0.05">
      <failure message="Always failing." type="java.lang.RuntimeException"><![CDATA[java.lang.RuntimeException: Always failing.
	at com.library.AlwaysFailTest.alwaysFail(AlwaysFailTest.java:20)
...
	at com.github.bazel_contrib.contrib_rules_jvm.junit5.JUnit5Runner.main(JUnit5Runner.java:39)
]]></failure>
    </testcase>
    <testcase name="doNothingTest" classname="com.library.AlwaysFailTest" time="0.03">
      <system-out><![CDATA[Hi there!
]]></system-out>
    </testcase>
  </testsuite>
</testsuites>
```

Co-authored-by: Vince Rose <vrose@confluent.io>

PR #328, opened 2025-03-10
fix: BazelJUnitOutputListener logging xml output on suite-level failures

amishra-u:
fixes: #327 

1. Updated code to propagate the test suite failure to each test case.
2. The test suite failure will be propagated to each child node, even if the test is not executed.
3. Additionally, this change fixes the bug introduced in #300, which incorrectly logged the test_suite as a testcase and logged incorrect values for the "tests" attribute in the <testsuite> node.

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BazelJUnitOutputListener.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BazelJUnitOutputListener.java
@@ -264,7 +264,6 @@ public class BazelJUnitOutputListener implements TestExecutionListener, Closeabl
         .collect(Collectors.toList());
   }
 
-
   @Override
   public void reportingEntryPublished(TestIdentifier testIdentifier, ReportEntry entry) {
     getResult(testIdentifier).addReportEntry(entry);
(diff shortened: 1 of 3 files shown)


fzakaria:
I wrote some tests that work on actual test classes rather than `TestData` -- maybe it would be good to continue in that direction.

The unit tests on TestData only take us so far; we need to write more "integration-style" tests running on real classes IMO since the JUnit data structure is kind of opaque.

shs96c:
@fzakaria, I've no objection to a very limited number of test cases like this, but they tend to be slow to run, require a lot more set up, and can be opaque to debug and understand. If we introduce them, I'd rather they were an absolute minority of tests for this kind of functionality.

fzakaria:
@shs96c i agree hollistically but the JUnit stuff is so opaque -- feels like starting in the reverse direction is good.
i.e. writing "integration style tests" and then taking that information to encode it as a unit test with the TestData mock.
I guess they can always be tagged and run in a separate step ;P

Anyways -- just my thoughts. As someone who likes to contribute this can help me understand if I'm breaking other consumers.

amishra-u:
@shs96c Can you please take a look

shs96c (approved):
LGTM, but could you please run `bazel run scripts:format`?

amishra-u:
@shs96c Can you please merge it. I have fixed formatting issue.

PR #331, opened 2025-03-15
Defer XML output file creation to avoid premature file generation

amishra-u:
#### Issue
The test runner creates an empty XML file when it terminates unexpectedly, such as when the JVM shuts down before the listener has a chance to write any output. The existing premature file creation interferes with Bazel’s fallback mechanism, which generates the XML file from test.log only if the output file is missing.

#### Solution
Use LazyFileWriter to delay file creation until the first write operation. This ensures that no empty file is created if the JVM shuts down before any content is written. If no output is generated, Bazel's [fallback logic](https://github.com/bazelbuild/bazel/blob/47e5113eb4b4a2eed3c2f1fd72a1fa9ad551a8cc/src/main/java/com/google/devtools/build/lib/exec/StandaloneTestStrategy.java#L854) can correctly generate the XML file from test.log, as it only activates when the output file does not already exist.


shs96c (approved):
This isn't thread-safe, but I don't think it needs to be. LGTM. Thank you for the PR!

PR #332, opened 2025-03-21
Ensure deps referenced in java_test_suite jvm_flags are passed to tests

thirtyseven:
https://github.com/bazel-contrib/rules_jvm/pull/303 introduced a regression where if a `java_test_suite` references a dep in its jvm_flags via `$(location)` or other Make variable, it will not be present in the deps of the generated `java_test` leading to analysis errors. This PR adds an example demonstrating the issue and suggests a possible fix.


thirtyseven:
Any feedback on this? Is there a better approach?

illicitonion (approved):
Seems reasonable! Thank you!

Issue #319, opened 2025-01-30
BazelJUnitOutputListener doesn't log methodName for @ParameterizedTest


amishra-u:
### How to reproduce?
1. Run any @Parameterized test inside rules_jvm package.
2. Check the generated text.xml file, the name in test case only contains index of the test and ignores the method name.
`test.xml` file for tests in MixedTest.java file

###  Expected
expected name to testMethodName and index
```
<testcase name="bootstrap[1] goGreek=alpha" classname="com.github.bazel_contrib.contrib_rules_jvm.comparative.MixedTest" time="0.03">
```

### Actual
But test name only contains index
```
<testcase name="[1] goGreek=alpha" classname="com.github.bazel_contrib.contrib_rules_jvm.comparative.MixedTest" time="0.03">
```

### Issue
Unit 5 internally treats @ParameterizedTest as a specialized form of a test template, storing it as an instance of [TestTemplateTestDescriptor](https://github.com/junit-team/junit5/blob/ee52c64336ad3a079d598bd1c3884bafbc285047/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestTemplateTestDescriptor.java) in the DiscoveryResult. For execution, TestTemplateTestDescriptor leverages [DynamicTestExecutor](https://github.com/junit-team/junit5/blob/ee52c64336ad3a079d598bd1c3884bafbc285047/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTestTask.java#L204), which triggers the dynamicTestRegistered method on registered listeners. As a result, [BazelJUnitOutputListener](https://github.com/bazel-contrib/rules_jvm/blob/8d835288e8d510f42a148f515a239cc32e165711/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/BazelJUnitOutputListener.java#L172) incorrectly classifies ParameterizedTest as a dynamic test.

christianladam:
For reference:

This can be resolved using `test --jvmopt="-Djunit.jupiter.params.displayname.default=\"[{index}] {displayName}=({arguments})\""` or by registering custom name formatters for repeated/parameterized junit5 tests.

That said, there are certain types that don't allow for that due to being junit4 (as you pointed out with JUnitParams).

PR #333, opened 2025-03-21
resolve issue 319, handling test method name for Parameterized tests

asbeane:
Attempts to resolve https://github.com/bazel-contrib/rules_jvm/issues/319

checks for the ordinal (in parameterized tests), and this case maps each test case to the method name and wraps each of these in their own test suite so the tests show up with each parameterized test as test case, and the method has a test suite.

Basically the hierarchy (for parameterized tests)

Test Class (also a test suiteA)
       Test case/method
             [1] params=...
             [2] params=....

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestCaseXmlRenderer.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestCaseXmlRenderer.java
@@ -34,17 +34,7 @@ class TestCaseXmlRenderer {
 
     TestIdentifier id = test.getId();
 
-    String name;
-    if (test.isDynamic()) {
-      name = id.getDisplayName(); // [ordinal] name=value...
-    } else {
-      // Massage the name
-      name = id.getLegacyReportingName();
-      int index = name.indexOf('(');
-      if (index != -1) {
-        name = name.substring(0, index);
-      }
-    }
+    String name = getTestName(test);
 
     xml.writeStartElement("testcase");
     xml.writeAttribute("name", escapeIllegalCharacters(name));
@@ -103,5 +93,33 @@ class TestCaseXmlRenderer {
     throwable.printStackTrace(new PrintWriter(stringWriter));
 
     xml.writeCData(escapeIllegalCharacters(stringWriter.toString()));
+  }
+
+  private String getTestName(TestData test) {
+    TestIdentifier id = test.getId();
+    // checking for the '[' as a proxy for the ordinal parameterized test.
+    // If there is some edge case not considered, here, it should still be okay, as it really just
+    // formats the test case name.
+    if (id.getParentId().isPresent() && id.getDisplayName().startsWith("[")) {
+      return getCustomDisplayName(id);
+    }
+    // this check assumes ParameterizedTest is dynamic, leaving this for now but may not need it.
+    // Edge case maybe when there is no parent (if ever possible)
+    if (test.isDynamic()) {
+      return id.getDisplayName(); // [ordinal] name=value...
+    } else {
+      String name = id.getLegacyReportingName();
+      int index = name.indexOf('(');
+      if (index != -1) {
+        name = name.substring(0, index);
+      }
+      return name;
+    }
+  }
+
+  private String getCustomDisplayName(TestIdentifier testIdentifier) {
+    TestIdentifier parent = testPlan.getTestIdentifier(testIdentifier.getParentId().get());
+    String methodName = parent.getDisplayName().replaceAll("\\(\\)", "");
+    return methodName + " " + testIdentifier.getDisplayName();
   }
 }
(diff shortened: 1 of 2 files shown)


shs96c (commented):
Thank you for the PR. This seems okay, but I've a couple of questions:

1. How does Gradle and/or Maven report this in their log files? We should be compatible with that, since those are widely used tools, and the XML files are sucked into CI runners to display results.
2. Can we please have a test for this? Without one, we're likely to regress this output as we continue to work in this area.

asbeane:
> We recently landed #351. Could you please LMK if this addressed your issue, or whether this PR still needs to be landed.

Hey @shs96c, apologies for being so slow. @mattnworb brought this back to my attention.
So I tested this today, and it looks like the previous change ([#351](https://github.com/bazel-contrib/rules_jvm/pull/351)) nicely added support for nested tests, but (from my own testing, and from what I can tell) did not add support for parameterized tests (which this PR does).

To answer your previous question - the intent of this PR is to reach compatibility with what maven does (at least to the best of my knowledge), specifically for Parameterized Tests.

I'll work on adding a test this per your suggestion - thank you for the reminder and follow-up @shs96c!

christianladam:
> > We recently landed #351. Could you please LMK if this addressed your issue, or whether this PR still needs to be landed.
> 
> Hey @shs96c, apologies for being so slow. @mattnworb brought this back to my attention. So I tested this today, and it looks like the previous change ([#351](https://github.com/bazel-contrib/rules_jvm/pull/351)) nicely added support for nested tests, but (from my own testing, and from what I can tell) did not add support for parameterized tests (which this PR does).
> 
> To answer your previous question - the intent of this PR is to reach compatibility with what maven does (at least to the best of my knowledge), specifically for Parameterized Tests.
> 
> I'll work on adding a test this per your suggestion - thank you for the reminder and follow-up @shs96c!

Yes #351 was done specifically to handle nesting and kept the previous behavior for parameterized tests alone on purpose. Technically one can specify a name value to the annotation to fix this but it's less than ideal behavior when one doesn't do that. The scenarios are:

1. no names in the annotations
  @ParameterizedTest

2. name set for the ParameterizedTest annotation that includes the template name
  @ParameterizedTest(name = "aTest({0})")

3. name set for the ParameterizedTest annotation that doesn't include the template name
  @ParameterizedTest(name = "just some customized string {0}")

Then have something like:
  @ValueSource(strings = {"one", "two", "three", "four"})
  void aTest(String value) {
    ...
  }

What's the desired behavior here? Should we always include the template in the grouping instead of the looking one step up for a class/nested-class?

christianladam:
For reference here are the XML formats for the same parameterized test class in bazel, gradle, and maven:

BAZEL
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="bootstrap(String)" timestamp="2025-07-25T23:19:11.359936Z" hostname="Macbook-Pro.local" tests="3" time="0.03" failures="0" errors="0" disabled="0" skipped="0" package="">
    <properties/>
    <testcase name="[2] goGreek=beta" classname="com.github.bazel_contrib.contrib_rules_jvm.junit5.ParameterisedTest" time="0">
      <system-out><![CDATA[beta
]]></system-out>
    </testcase>
    <testcase name="[1] goGreek=alpha" classname="com.github.bazel_contrib.contrib_rules_jvm.junit5.ParameterisedTest" time="0.01">
      <system-out><![CDATA[alpha
]]></system-out>
    </testcase>
    <testcase name="[3] goGreek=gamma" classname="com.github.bazel_contrib.contrib_rules_jvm.junit5.ParameterisedTest" time="0">
      <system-out><![CDATA[gamma
]]></system-out>
    </testcase>
  </testsuite>
</testsuites>
```

GRADLE: https://github.com/bazel-contrib/rules_jvm/blob/main/comparative-tests/build/test-results/test/TEST-com.github.bazel_contrib.contrib_rules_jvm.comparative.ParameterisedTest.xml

MAVEN (surefire): https://github.com/bazel-contrib/rules_jvm/blob/main/comparative-tests/target/surefire-reports/TEST-com.github.bazel_contrib.contrib_rules_jvm.comparative.ParameterisedTest.xml

The main differences include the bazel XML having an outer <testsuites>, maven and bazel sharing the same system-out data while gradle has a list outside of the testcases as well as having a system-err line, and the naming being:

bazel [index] parameterName=value
gradle [index] value
maven methodName(type)[index]

christianladam:
After looking into this more, repeated and parameterized tests are using the default naming scheme defined in the junit5 code. There's multiple ways to override the default behavior including doing things like creating your own NameFormatters and registering yours to override the default ones. It's probably just easier to go ahead and add the following to your .bazelrc to test:
```
test --jvmopt="-Djunit.jupiter.params.displayname.default=\"[{index}] {displayName}=({arguments})\""
```

Which would make the output go from:
[1] goGreek=alpha
to:
[1] bootstrap(String)=(alpha)

For other types one can always register custom formatters. See https://docs.junit.org/current/user-guide/#writing-tests-display-names for more details.

asbeane:
Cool, thank you @christianladam! Taking a look.

PR #335, opened 2025-03-27
Handle duplicate test names in report

amishra-u:
JUnitParams generates report names based on parameter values rather than parameter types, which can result in duplicate test names in the XML output. This situation arises when two test methods have identical names but differ in parameter types; however, their string representations may be the same (e.g., Integer 1 and Long 1).

Added logic to ignore tests with duplicate names in test report names.

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestSuiteXmlRenderer.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestSuiteXmlRenderer.java
@@ -10,7 +10,9 @@ import java.time.Duration;
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Locale;
+import java.util.Set;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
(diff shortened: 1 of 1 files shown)


amishra-u:
@illicitonion can you please review?

shs96c:
#351 recently landed. I'm pretty sure that we still need this PR, but could you please confirm? If we do, I'll be happy to merge this.

Please also accept my apologies for the very slow review.

amishra-u:
@christianladam this is talking about junitparmsrunner. This flag works for Junit5 Parameterized test.
https://github.com/Pragmatists/JUnitParams/blob/master/src/main/java/junitparams/JUnitParamsRunner.java

christianladam:
Do you want it to ignore duplicates? How would one know which result its actually showing? It feels like it'd be better to change the format of the JUnitParams stuff to handle it better. Since JUnitParamsRunner is junit4, do you know what the behavior is when using the standard junit4 test runner provided by bazel? Ideally we'd be handling things the same in both cases. If it's having this issue in both cases, then ideally there'd be a solution that works for both where potentially the JUnitParams formatting via https://github.com/Pragmatists/JUnitParams/blob/master/src/main/java/junitparams/naming/MacroSubstitutionNamingStrategy.java could be overridden or configured. If it's working fine with the standard junit4 runner but not with the logic in rules_jvm, then that should be fixed.

PR #339, opened 2025-04-21
chore: Avoid modifying deps argument

honnix:
It is not great to modify the passed-in `deps` argument because users might use the same variable in a later macro/rule invocation but find it has been tainted.


honnix:
I'm not familiar enough with the codebase to find a good place to test this change, so please help me understand where I could do that.

PR #347, opened 2025-06-18
feat: Optionally respect java_export during resolution

### LINKED ISSUE #344: FR: Support `java_export` targets with the gazelle plugin
A key mechanism for Bazel repos to interact with the wider java ecosystem is the `java_export` rule. This allows bazel-built classes to be exported as regular jars to Maven repos. It would be very useful if the java gazelle plugin understood the semantics of `java_export` and properly generated build files that accounted for it.

The key thing is that if a target is included in a `java_export` it should never be referred to directly, except by other targets that contribute to the same `java_export` target. This helps prevent things like split packages, and avoids having duplicate classes appear in multiple different jars (a source of "classpath hell")

There are two places where this kind of functionality is useful:

1. When migrating from maven or gradle, placing a `java_export` at the top of each project in a multi-project repo would maintain the same semantics and "shape" of jars being exported to Maven repos.
2. Within a monorepo, this would allow portions to be shared with projects outside of the monorepo. 

An example of this second case is [Selenium](https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/), which has many `java_export` targets within the same repo. Maintaining the build files by hand to ensure that classes only appear in only one jar is tedious and error prone.

## Key requirements

1. A class that appears in a `java_export` should not appear in any other `java_export` target in the tree.
2. Targets "outside" of the `java_export` should only ever refer to the `java_export`, not the specific `java_library` targets that are used "within" the `java_export`
3. Within a `java_export` target, fine-grained targets should still be used.

There is an important exception to the first requirement: if a user deliberately adds the same target as a dependency to two different `java_export` targets, then the same classes may appear in more than one exported jar.

## Expected usage

I would not expect Gazelle to create the `java_export` target. My expectation would be that I would change an existing `java_library` target to `java_export`, or would create a new `java_export` target that contains dependencies on targets generated by gazelle.

Bearing in mind the "key requirements", fine-grained build targets would still be used as much as possible, allowing large subtrees to still build efficiently.

### Gradle/Maven multi-project project

```
+- module1
    +- BUILD.bazel
    +- src/main/java/
        +- com/example/
           +- BUILD.bazel
           +- foo/
               +- BUILD.bazel
+- module2
    +- BUILD.bazel
    +- src/main/java/
        +- com/example/bar/
               +- BUILD.bazel
         +- com/example/baz/
               +- BUILD.bazel
```

In this example, there is a `java_export` in `module1/BUILD.bazel` and `module2/BUILD.bazel`. These include everything in the directories below them.

Build files in `module1` would only refer to the `module2` `java_export` target if they needed classes from `com.example.bar` or `com.example.baz`. Similarly, build files in `module2` would only refer to the `java_export` target in `//module1` if they needed classes from `com.example` or `com.example.foo`.

### Bazel-built monorepo

```
+ java/src/com/example
   +- BUILD.bazel
   +- foo/
       +- BUILD.bazel
   +- bar/
       +- BUILD.bazel
   +- baz/
       +- BUILD.bazel
```

This is conceptually identical to the gradle/maven example.

`//java/src/com/example:BUILD.bazel` could look like:

```starlark
java_export(
    name = "export-foo",
    deps = [
        ":example",
        "//java/src/com/example/foo",
    ],
    maven_coordinates = "....",
)

java_export(
    name = "export-others",
    deps = [
        "//java/src/com/example/bar",
        "//java/src/com/example/baz",
    ],
    maven_coordinates = "....",
)

java_library(
    name = ":example",
    srcs = glob(["*.java"]),
)
```

Again, direct dependencies to `java_library` targets from (eg) `//java/src/com/example/bar` to `//java/src/com/example/baz` would be allowed, but if targets within either of these bazel packages needed access to classes in `com.example.foo`, they'd need to refer to `//java/src/com/example:export-foo`

## I want a pony 🦄

The obvious issue with this is that tests would depend on the `java_export` targets, which would to inefficient builds, firstly because more of the source under test might need to be compiled before the test can start, and because the coarse-granularity of the `java_export` targets would cause more tests to run than would be possible with a very fine-grained build graph. If I could have a pony, it would be helpful if test targets could (optionally) directly access the fine-grained targets at all times (though this might lead to other problems)

blorente:
Implement support for `java_export`s. See #344 for a description of the problem.

Constraints we have to work with:
- We make no attempt to modify `java_export` targets. They are meant to be hand-crafted by the user.
- `java_export` uses an aspect to accumulate transitive dependency information about its dependencies. We need this transitive dependency information, and the Gazelle index doesn't give it to us. Therefore, we have to store it out of band, in `java_export_index.go`. The community seems to agree that the Gazelle API doesn't allow for this ([GH Issue](https://github.com/bazel-contrib/bazel-gazelle/issues/2123), [Slack discussion](https://bazelbuild.slack.com/archives/C01HMGN77Q8/p1750412735065129)).
- It's possible for a `java_export` to depend on a package that is _not_ in a subdirectory. Therefore, we can't rely on the `Imports` or `Embeds` phase to operate on the transitive dependency information, because some relevant packages may not have been analized by the time we get to a `java_export`. There is a test case covering this. As a result, the `java_export_index` is split into two phases:
  - The first phase happens during `GenerateRules`, we note down information about all `java_exports`, as well as information to resolve their transitive dependencies.
  - At the end of `GenerateRules`, in the `DoneGeneratingRules` hook, we traverse the information we've collected to figure out which Java packages are accessible from a given `java_export`.
  - The second phase happens during `Resolve`, where we use the data we've collected to resolve to a `java_export`, if possible.
- This feature can only be turned on via the `# gazelle:java_resolve_to_java_exports` directive. This directive can only be set once, and it has to be set at the root of the repository.
  - It can only be set once to avoid edge cases where a package opts out of resolving to `java_exports`, but it is inside a `java_export` itself.
  - It can only be set at the root of the repository because we need to run Gazelle over the entire repository, to make sure we capture all the transitive dependencies.

[This commit](https://bazelbuild.slack.com/archives/C01HMGN77Q8/p1750412735065129) is an example of running this Gazelle extension in the Selenium repository after massaging the BUILD files a bit to fit the constraints of `java_exports`.

Closes #344


java/gazelle/testdata/java_export_targets/module1/src/main/java/com/example/module1/BUILD.out
```diff
@@ -0,0 +1,18 @@
...
+)
+
+java_library(
+    name = "module1",
+    srcs = [
+        "Module1.java",
+        "Module1InterfaceToFoo.java",
+    ],
+    visibility = ["//:__subpackages__"],
+    exports = ["//module1/src/main/java/com/example/module1/foo"],
```
shs96c, line 16: You don't need this `exports` for the dep on `module_export` to work as it should

java/gazelle/testdata/java_export_targets/module1/src/main/java/com/example/module1/foo/BUILD.out
```diff
@@ -0,0 +1,7 @@
+load("@rules_java//java:defs.bzl", "java_library")
+
+java_library(
+    name = "foo",
+    srcs = ["Module1Foo.java"],
+    visibility = ["//:__subpackages__"],
```
shs96c, line 6: This will need to be constrained to prevent people accessing the target by accident.

blorente: done now.

stevebarrau (commented):
I've tested this at a small scale, and it works as advertised. Thanks, @blorente!

One quick question:

> This feature can only be turned on via the # gazelle:java_resolve_to_java_exports directive.

It is enabled by default, right?

I have very lightly tested the case where multiple java_export targets transitively depend on the same, non-exported java_library. Something like:

```
a -> exported_of_b -> c  
a -> exported_of_d -> c
```

If I understand correctly, this isn't advised by Java and should trigger an error, prompting the user to wrap c in a java_export target. @shs96c will know more, I only know such duplication cause headaches down the lines for Gazelle resolving classes in packages in split JARs.

With the existing work in this PR, it made `exported_of_d` depend on `exported_of_b` (and not on `c`) which, albeit arbitrary, is correct. Do you have any thoughts on this?

stevebarrau (approved):
LGTM. Thanks for putting this together.

Happy maven publishing!

PR #350, opened 2025-07-23
feat: Allow building a version of the plugin that embeds the javaparser server

blorente:
Thus allowing it to be "statically linkable".

The user-facing UX has two options:
- To use the command line, like this: `bazel build //:<your-gazelle> '--@contrib_rules_jvm//java/gazelle:embed_server=True'`
- To create a transition, like the one we use for testing in `java/gazelle/testdata/gazelle_build_mode.bzl`.

Caveats and implementation notes:
- We always need a java binary at runtime. This can be specified via `JAVA_HOME`, or a new environment variable, `GAZELLE_JAVA_BIN`. We don't strictly need the new env var, we should be able to always use `JAVA_HOME`, but I thought it would be nice to have a variable explicitly for the Java plugin, just in case it differs from your runtime environment.
- We implement the embedding directly in `servermanager.go`, by conditionally compiling one of two versions of the library.
- We need to do some hacks for testing, because we need to wrap the actual gazelle binary, and the tests do not run the gazelle binary from Bazel's sandboxed working directory.
- When running the embedded server, we materialize the javaparser server jar 100% of the time. We could be smarter about this, but the server is small enough that I don't think it matters.

### DIFF
--- a/java/gazelle/private/servermanager/BUILD.bazel
+++ b/java/gazelle/private/servermanager/BUILD.bazel
@@ -10,17 +10,17 @@ go_library(
         "//conditions:default": [
             "insource-server.go",
         ],
-    }),
+    }),  # keep
     data = select({
         "//java/gazelle:embedded_server": [],
         "//conditions:default": ["//java/src/com/github/bazel_contrib/contrib_rules_jvm/javaparser/generators:Main"],
-    }),
+    }),  # keep
     embedsrcs = select({
         "//java/gazelle:embedded_server": [
             ":javaparser_deploy_jar",
         ],
         "//conditions:default": [],
-    }),
+    }),  # keep
     importpath = "github.com/bazel-contrib/rules_jvm/java/gazelle/private/servermanager",
     visibility = ["//java/gazelle:__subpackages__"],
     deps = [


java/gazelle/private/servermanager/BUILD.bazel
```diff
@@ -12,3 +28,10 @@ go_library(
         "@rules_go//go/runfiles",
     ],
 )
+
+genrule(
```
stevebarrau, line 32: Should we use `load("@aspect_bazel_lib//lib:copy_file.bzl", "copy_file")` here?

java/gazelle/private/servermanager/BUILD.bazel
```diff
@@ -2,8 +2,24 @@ load("@rules_go//go:def.bzl", "go_library")
...
+    data = select({
+        "//java/gazelle:embedded_server": [],
+        "//conditions:default": ["//java/src/com/github/bazel_contrib/contrib_rules_jvm/javaparser/generators:Main"],
+    }),
+    embedsrcs = select({
+        "//java/gazelle:embedded_server": [
+            ":javaparser_deploy_jar",
+        ],
+        "//conditions:default": [],
+    }),
```
stevebarrau, line 22: Does gazelle-go 1. like 2. generate these selects?

java/gazelle/private/servermanager/embedded.go
```diff
@@ -0,0 +1,82 @@
...
+	runnerPath := filepath.Join(tmpdir, "runjavaparser")
+	runnerTemplate := unixRunerTemplate
+
+	if runtime.GOOS == "windows" {
+		runnerPath += ".exe"
+		runnerTemplate = windowsRunnerTemplate
+	}
+
+	javaBin :=
+		os.Getenv("GAZELLE_JAVA_BIN")
```
stevebarrau, line 51: Should this GAZELLE_JAVA_BIN be:

1. a const so it carries the doc string?
2. renamed to a more verbose yet systemic GAZELLE_JAVA_JAVAHOME

java/gazelle/private/servermanager/embedded.go
```diff
@@ -0,0 +1,82 @@
...
+	if javaBin == "" {
+		javaHome := os.Getenv("JAVA_HOME")
+
+		if javaHome == "" {
+			panic("Could not find GAZELLE_JAVA_BIN or JAVA_HOME. When running on embedded mode the Gazelle extension for Java requires a java executable. Please set it using GAZELLE_JAVA_BIN or disable the extension.")
+		}
+
+		javaBin = filepath.Join(javaHome, "bin", "java")
+	}
```
stevebarrau, line 60: Could the be a find **first** value in the list of getenv of `[]{"GAZELLE_JAVA_JAVAHOME", "JAVA_HOME"}`, and error check outside the loop?

java/gazelle/private/servermanager/insource-server.go
```diff
@@ -0,0 +1,37 @@
...
+func (m *ServerManager) locateJavaparser(jvmFlags []string) (string, error) {
+	rf, err := runfiles.New()
+	if err != nil {
+		return "", fmt.Errorf("failed to init new style runfiles: %w", err)
+	}
+
+	javaparserPath := javaparserLocation()
+	if runtime.GOOS == "windows" {
+		javaparserPath += ".exe"
+	}
```
stevebarrau, line 23: This exists twice, could we reasonably DRY it?

java/gazelle/private/servermanager/BUILD.bazel
blorente: No, it definitely doesn't like these selects. I think however that some selects are necessary if we want to avoid compiling the non-embedded version when we compile the embedded one, and this is the lowest-impact place I could find for them.

java/gazelle/private/servermanager/embedded.go
blorente: Yep, definitely can be :D

java/gazelle/private/servermanager/insource-server.go
blorente: I'm not sure, because while the condition is the same as `embedded.go`, the effect is different. In embedded.go, we definitely want a `.bat` file, because it's a script and not an executable. Happy to store a boolean in `ServerManager`, but it doesn't feel like something the servermanager should care about.

java/gazelle/private/servermanager/embedded.go
blorente: Yep! Done :D Great idea.

java/gazelle/private/servermanager/BUILD.bazel
blorente: Good call! Done :)

MODULE.bazel
```diff
@@ -21,6 +21,7 @@ bazel_dep(name = "rules_go", version = "0.52.0")
 bazel_dep(name = "rules_java", version = "7.12.2")
 bazel_dep(name = "rules_jvm_external", version = "6.7")
 bazel_dep(name = "rules_proto", version = "7.0.2")
+bazel_dep(name = "aspect_bazel_lib", version = "2.20.0")
```
shs96c, line 24: As we're a dependency that people will pull in, can we keep this version as low as possible to avoid forcing users to upgrade `aspect_bazel_lib` unless they want to

java/gazelle/private/servermanager/embedded.go
```diff
@@ -0,0 +1,86 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+// The embedded javaparser server (and runner script) will be materialized at runtime every time we run gazelle.
+// This is not a problem for now, as the javaparser server is tiny.
+// However, we may want to get clever and materialize to a more permanent location in the future if the server becomes significant.
```
shs96c, line 14: Do we make an effort to delete it as we exit the app?

shs96c (approved):
LGTM

PR #351, opened 2025-07-23
fix: support nested classes properly

christianladam:
Previously if you had deeply nested classes the BazelJUnitOutputListener would throw an exception. In JUnit5 TestExecutionListeners are not expected to throw exceptions and any exception just gets swallowed and logged as a warning. This would cause cases where an exception was thrown to output no test results and result in a successful test run even though there was a failure in the listener itself. The new logic attempts to get the closest parent segment that is a class type. This doesn't match the way things like gradle will output test XML but solves the problem with nested classes while keeping the output similar.

In terms of output this actually fixes an issue that is visible in the NestedClassesTest today, where the Second nested classes function result gets grouped into the outer class while the static First classes function result appears in a separate group as shown below:

```
NestedClassesTest
  shouldBeExecuted
  shouldBeExecuted
NestedClassesTest$First
  shouldBeExecuted
```

With the changes the grouping is fixed as shown below:

```
NestedClassesTest
  shouldBeExecuted
NestedClassesTest$First
  shouldBeExecuted
NestedClassesTest$Second
  shouldBeExecuted
NestedClassesTest$Second$Third
  shouldBeExecuted
NestedClassesTest$Second$Third$Fourth
  shouldBeExecuted
NestedClassesTest$Second$Third$Fourth$Fifth
  shouldBeExecuted
```

These groups appear unordered in UI like IntelliJ which isn't ideal, but these changes closely align with how things are done today, while avoiding just silently passing with no test output if you take this version of the NestedClassesTest and run it using the previous BazelJUnitOutputListener code.

### DIFF
--- a/comparative-tests/build/test-results/test/TEST-com.github.bazel_contrib.contrib_rules_jvm.comparative.NestedClassesTest$First.xml
+++ b/comparative-tests/build/test-results/test/TEST-com.github.bazel_contrib.contrib_rules_jvm.comparative.NestedClassesTest$First.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuite name="com.github.bazel_contrib.contrib_rules_jvm.comparative.NestedClassesTest$First" tests="1" skipped="0" failures="0" errors="0" timestamp="2025-07-23T17:23:19" hostname="MacBook-Pro.local" time="0.0">
+  <properties/>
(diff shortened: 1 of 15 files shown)


christianladam:
Fixed the issue in the test output test where the children weren't being picked up because they weren't actually being added to the test plan. This is due to the original code always adding the groups but not necessarily adding tests. With the new code we only add the groups when there's actually a test.

PR #352, opened 2025-08-01
chore: add parameterized and repeated tests to the junit5 test target

christianladam:
Also generated the comparative xml for the repeated test


christianladam:
We had ParameterisedTest in the comparative tests but not in the junit5 tests. Adding it to the junit5 tests to enable developers to easily run and compare the xml output from the Bazel runner vs the gradle/maven xml. Also added a RepeatedTest and the comparative XMLs that generates.

christianladam:
Pushed the formatting fixes.

PR #361, opened 2025-08-22
fix: kotlin names and cucumber tests

christianladam:
There are two main changes in this PR.

The first is to support special cases like cucumber tests where the method itself doesn't have a Source object. This is because the tests come from the feature files rather than an actual source file. Instead of just skipping such tests we find the first parent with a source. This allows running individual tests via intelliJ to still work when the cucumber test class is being run.

The second change allows regex specific characters to work with the PatternFilter. This is required when kotlin names functions like `method() succeeds` and `method() fails` which currently causes duplicates in the XML report. Instead of massaging the name to remove everything after a ( is found, we now handle such characters properly in the filter code. This change does exclude . from the escaping because intelliJ separates class and methods when running individual tests with a . instead of #. This means we will potentially over execute tests but it's better to ensure the desired test is run than to not run anything.

Note: If kotlin support is added the FilteringTests should be updated to include special character function names to improve the test coverage.


shs96c (approved):
LGTM. Thank you!

PR #365, opened 2025-09-08
Bump `rules_jvm_external` to 6.8 and switch resolver

shs96c:
With this release, the default resolver can handle BOM files. It's a better choice for us, since the maven dependency resolution mechanism isn't great.


MODULE.bazel
```diff
@@ -143,7 +143,6 @@ maven.install(
     fail_if_repin_required = True,
     generate_compat_repositories = True,
     lock_file = "//:contrib_rules_jvm_deps_install.json",
-    resolver = "maven",
```
gibfahn, line 146: So to be clear, this switches us back to coursier as the resolver right?

shs96c: Yes, back to `coursier` now that it supports boms

PR #371, opened 2025-09-22
report assertion errors as 'failures'  and other runtime exceptions as 'errors', consistent with Junit terminology

subbu171:
**What?**

Currently, assertion failures (e.g., AssertionError) are being reported as errors, and unexpected exceptions (e.g., NullPointerException, IOException, etc.) are being reported as failures.

This is inverted relative to JUnit’s standard terminology.

**JUnit convention**

Failures → Assertion failures (AssertionError) indicating the test ran but produced an unexpected result.

Errors → Unexpected exceptions, infrastructure issues, or anything other than an AssertionError.

Reference: 
JUnit 5 Legacy Reporter
Logs AssertionError as FAILURE and everything else as ERROR.
Reference: [JUnit Legacy XMLReportWriter.java (line 444)](https://github.com/junit-team/junit-framework/blob/main/junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/xml/XmlReportWriter.java#L444)

**Fix**

Update error classification logic to correctly map:

AssertionError → failure

Any other exception → error

**Impact**

Test reports will now align with JUnit expectations.

Downstream tooling and dashboards that parse JUnit XML or test results will correctly distinguish between test failures (assertion mismatches) and infrastructure errors (unexpected exceptions).

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestData.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestData.java
@@ -80,10 +80,6 @@ class TestData {
       return false;
     }
 
-    if (result.getStatus() == TestExecutionResult.Status.ABORTED) {
-      return true;
-    }
-
     return result.getThrowable().map(thr -> (!(thr instanceof AssertionError))).orElse(false);
   }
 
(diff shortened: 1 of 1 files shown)


shs96c (changes requested):
One small nit, but this looks great. Thank you for catching this!

Issue #316, opened 2025-01-10
java_test_suite has trouble finding srcs specified as a filegroup


programmablemike:
I’m having an issue in one of my projects with the `bazel-contrib/rules_jvm` package `java_test_suite` where I specify the `srcs` using a `filegroup`, but the source files are never found. However, using a `glob(["src/test/java/**/*.java"])` or manually specifying the list of source files does work.

For reference I’m using Bazel 8.0.0 (via Bazelisk) and the following MODULE dependencies:

```
bazel_dep(name = "rules_java", version = "8.6.3")
bazel_dep(name = "rules_jvm_external", version = "6.6")
bazel_dep(name = "contrib_rules_jvm", version = "0.27.0")
```

I put together an example repository to demonstrate the issue here: https://github.com/programmablemike/issue-bazel-java-test-suite-filegroup/

sitaktif:
Turns out the problem is that `java_test_suite` is a macro that ends up iterating on the `srcs`, expecting strings. The `filegroup` can't be introspected by the macro at the time it's expanded.

Only thing I can suggest is to assign the glob to a variable and use it both in the `filegroup` and in the `java_test_suite` call.

PR #378, opened 2025-10-01
[gazelle] Add java_include_binary directive

tjoneslo:
This is the first part of Issue #316, where it was requested that a new directive be added to skip adding the `java_binary` rule to specific directories.

Add tests to verify this works as expected.


shs96c (approved):
LGTM

java/gazelle/javaconfig/config.go
```diff
@@ -113,6 +118,7 @@ type Config struct {
 	extensionEnabled                                   bool
 	isModuleRoot                                       bool
 	generateProto                                      bool
+	includeBinary                                      bool
```
H5-O5, line 121: the type is bool. Will it work? the purpose is to "skip adding the `java_binary` rule to specific directories" so I think map[string]bool is needed to remember 'what paths are excluded'.

tjoneslo: If you want to skip adding `java_binaries` in specific directories, you need to add the directive to the build file in those directories. So, for example:  if you have a directory structure like this:

`src/main/java/com/example/no-binary/add-binary`  you would need to have a `BUILD` file in the `no-binary` directory with the ` #gazelle:java_include_binary false` and then in the `add_binary` directory would have the `# gazelle:java_include_binary true` directive. Gazelle correctly interprets the multiple settings for the different directories.

java/gazelle/javaconfig/config.go
```diff
@@ -149,6 +155,7 @@ func New(repoRoot string) *Config {
 		extensionEnabled:       true,
 		isModuleRoot:           false,
 		generateProto:          true,
+		includeBinary:          true,
```
sitaktif, line 158: How do you feel about renaming it to `generateBinaryTarget` / `generate_binary_target`?

As a user, "include" makes me think of file inclusion, and I feel the suggestion above is a bit more self-documenting. Also, we have a `generateProto` above, so it feels a tad more consistent. Just a suggestion :-)

sitaktif (approved):
LGTM, but see if you like my renaming suggestion

PR #379, opened 2025-10-02
Fix the Windows CI builds

shs96c:


stevebarrau (approved):
Thanks for investing in the Windows support.

PR #380, opened 2025-10-07
[gazelle] update gazelle readme to make tables readable

tjoneslo:
The table syntax for the Github markdown is much more limited than the table syntax for the generated docs. So fix the Gazelle Readme so the docs appear as readable. 

Add the  sourceset root, strip resources prefix and include binary directives. in the documentation.


stevebarrau:
Could we use .rst and have tables that look like [this](https://github.com/bazel-contrib/bazel-gazelle?tab=readme-ov-file#update-repos)? Note how this spans both columns.

Screenshot 2025-10-07 at 14 39 27

PR #382, opened 2025-10-21
Expose portions of the java_junit5_test macro logic via //java:defs.bzl

jeffmace:
# Summary

Take portions of the `java_junit5_test` macro and expose it as a method and constant via `//java:defs.bzl`.

## Problem

The `java_junit5_test` macro has some internal logic for preparing `java_test` to work with `JUnit5Runner`. Creating a test runner that wraps `JUnit5Runner` would have to duplicate that logic in order to expose the correct functionality.

## Solution

Take the logic out of the macro and place it into a function that can be called by any macro when creating the `java_test` target.

## Testing

Added test cases to ensure the logic is available to be imported and nominally returns the correct value. The test does not cover all branches of the jvm_flags logic for `java_junit5_test`.

### DIFF
--- a/examples/tests_and_lints/src/test/com/github/bazel_contrib/contrib_rules_jvm/examples/tests_and_lints/java_junit5_test_test_suite.bzl
+++ b/examples/tests_and_lints/src/test/com/github/bazel_contrib/contrib_rules_jvm/examples/tests_and_lints/java_junit5_test_test_suite.bzl
@@ -1,5 +1,6 @@
 load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
-load("@contrib_rules_jvm//java:defs.bzl", "java_junit5_test")
+load("@contrib_rules_jvm//java:defs.bzl", "java_junit5_test", "junit5_jvm_flags")
+load("@rules_java//java:java_test.bzl", "java_test")
 
 TargetInfo = provider(
     doc = "Information relating to the target under test.",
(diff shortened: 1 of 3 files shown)


mzeren-vmw (approved):
This LGTM, and test works for me.

shs96c:
The main thing that the `java_junit5_test` does is set the `main_class`. I'm not sure what the value of allowing that key detail to be overridden gives. Could you please explain a little more about the use case for this change?

mkosiba:
If you're changing `main_class` wouldn't you need to change the `runtime_deps` as well?

My $0.02 is that from a maintainability perspective I don't love adding a `main_class` argument as it's implying `java_junit5_test` now supports all sorts of runners, but that's not really the case. There is no formal API contract, there is no test coverage. If you subclass JUnit5Runner it'll likely work, if you don't it'll likely not work. So in effect we'd be adding an untested and effectively unsupported feature (unsupported in the sense that whenever rules_jvm changes in a way that breaks your use case, it won't be considered a bug/breaking change and you'll need to update your subclass/fork/etc.. to match).

The alternative is for you to copy the `java_junit5_test` macro into your repo, change the main class and use that. That use gives you pretty much all the guarantees you're already getting. The only difference is that with this PR, if the `jvm_flags` computation changes then you wouldn't need to update your fork.

This makes me think that perhaps moving the `java_test` kwargs computation into a helper function (that you can then use) or changing from `java_test` to a configurable factory function is a better way forward in this situation.

PR #383, opened 2025-10-21
Update java_test_suite to support custom test_generators instead of the fixed TEST_GENERATORS

jeffmace:
Using a fixed TEST_GENERATORS value prevents adding new test runners or customizing the `java_test` definitions for JUnit4 and JUnit5. This allows the caller to pass in functions that will define the corresponding `java_test` targets.

Example:

```
def my_custom_junit4_test(name, **kwargs):
    pass

def my_custom_junit5_test(name, **kwargs):
    pass

java_test_suite(
    name = name,
    test_generators = {
        "junit4": my_custom_junit4_test,
        "junit5": my_custom_junit5_test,
    }
)
```

### DIFF
--- a/examples/tests_and_lints/MODULE.bazel
+++ b/examples/tests_and_lints/MODULE.bazel
@@ -47,3 +47,8 @@ use_repo(
     maven,
     "maven",
 )
+
+register_toolchains(
+    "@@bazel_skylib//toolchains/unittest:cmd_toolchain",
+    "@@bazel_skylib//toolchains/unittest:bash_toolchain",
+)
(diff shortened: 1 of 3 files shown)


java/private/java_test_suite.bzl
```diff
@@ -80,7 +85,7 @@ def java_test_suite(
         package = package,
         define_library = _define_library,
         # Default to bazel's default test runner if we don't know what people want
-        define_test = _TEST_GENERATORS.get(runner, _define_junit4_test),
+        define_test = test_generators.get(runner, _define_junit4_test),
```
mzeren-vmw: Hm. Perhaps this should be:
```
test_generators.get(runner, _TEST_GENERATORS.get(runner, _define_junit4_test))
```
This way users have access to the default generators by just leaving them out of `test_generators`.

mzeren-vmw: Jeff pointed out offline that the default generators _are_ exported. See https://github.com/bazel-contrib/rules_jvm/blob/main/java/defs.bzl#L40-L41

PR #384, opened 2025-10-26
Update the Gazelle Java plugin to handle `java_maven_install_file` directives in nested directories

jeffmace:
Create a `maven.Resolver` for each instance of `java_maven_install_file` and use that resolver for all files under that directory. There aren't any changes to the default behavior; but nested `java_maven_install_file` declarations will now work as expected.

**Background**
The Gazelle plugin parses `java_maven_install_file` directives in nested directories but doesn't use the file when resolving dependencies. This is a problem for projects that use different Maven repositories for different parts of the repository. The top-level `maven_install.json` file may have incomplete or extra dependencies than the one configured for the lower level directory.

**Testing**
The unit tests have been updated to reflect the new interface. Integration tests have been added to confirm the updated behavior for an invalid `maven_install.json` file.

### DIFF
--- a/java/gazelle/resolve.go
+++ b/java/gazelle/resolve.go
@@ -283,27 +283,29 @@ func (jr *Resolver) resolveSinglePackage(c *config.Config, pc *javaconfig.Config
 		return l
 	}
 
-	if mavenResolver, err := jr.lang.MavenResolverForInstallFile(pc.MavenInstallFile()); err != nil {
+	mavenResolver, err := jr.lang.MavenResolverForInstallFile(pc.MavenInstallFile())
+	if err != nil {
 		jr.lang.logger.Fatal().Err(err).Msg("error creating Maven resolver")
-	} else {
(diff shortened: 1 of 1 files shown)


shs96c:
Apologies for the slow review. Looking now.

java/gazelle/testdata/maven_with_multiple_repositories/maven_install.junit.json
shs96c: This should also be a v2 lock file

jeffmace:
> I've a feeling that this isn't going to work when the lock file isn't in the repo root. That's because [MavenInfoFile()](https://github.com/bazel-contrib/rules_jvm/blob/main/java/gazelle/javaconfig/config.go#L243-L245) also looks for the lock file against the repo root and not the directory where the directive was discovered. I think you'll need to store the full path to the lock file and not just the path given by the directive.

I've updated the test case to show that providing the full path to the lock file will work. I believe this matches the current behavior because there aren't any changes to `maven.NewResolver`.

PR #385, opened 2025-10-28
Add a test case for paths like `src/main/com/example/test` and update IsTestPackage to handle it.

jeffmace:
Paths like this were previously added to `java_test_suite` instead of `java_library`.


jeffmace:
The test failures seem to be inconsistent. Have you seen them on other PRs?

shs96c:
The Windows jobs are flaky. It's most frustrating. I'll give them a nudge

PR #390, opened 2025-11-13
Fix pmd_test to use short_path instead of path for sources

styurin:
From https://bazel.build/rules/lib/builtins/File#path:

> Use the `short_path` for the path under which the file is mapped if it's in the runfiles of a binary. 

Fixes an issue with `pmd_test` not finding sources produced by other targets.


shs96c (approved):
LGTM. Thank you!

PR #391, opened 2025-11-15
Fix shutdown hooks for JUnit5 for JDK24+.

evanloriot:
Previously, all tests would exit with `System.exit or Runtime.exit was called!` because the shutdown hook was not removed at test teardown.


shs96c:
Can you please run `./tools/format.sh`? That should fix the formatting build

PR #394, opened 2025-11-28
feat(java/gazelle): scaffold class-level dependency resolution

shs96c:
Currently, the Java Gazelle plugin relies solely on package-level indexing for dependency resolution. This is inefficient for split packages in large corporate repositories, often requiring manual #gazelle:resolve directives.

This change introduces the architecture for class-level resolution:

Updates the resolution input pipeline to carry class-level import data alongside package data. Extends the Maven resolver to support querying by class name (with package fallback). Updates the resolution logic to prioritize precise class matches over broad package matches.


shs96c:
Is this with the most recent commits, or before I added them yesterday?

PR #399, opened 2025-12-06
Add a directive to enable/disable generating resources targets

jeffmace:
The new functionality from #370 may conflict with existing implementations. This provides an option to disable it globally or for specific directories.

# Example
```
# Disable generated targets for the resources directories
# gazelle:java_generate_resources false
```

# Testing Done

Added a new set of examples to the `java/gazelle/testdata/resources_custom_sourceset` directory.


shs96c (approved):
LGTM. Thank you for this!

PR #420, opened 2026-02-12
[gazelle] Add `java_generate_proto_services` directive

shs96c:
Allow users to disable `java_grpc_library` rule generation when `proto_library` rules with services are present, while still generating `java_proto_library` and `java_library` rules.


stevebarrau (approved):
This is going to be useful for Wire and ServiceTalk users.

Issue #410, opened 2026-01-29
Support for other test reporting formats


amilbourne:
Apologies if there is a simple workaround that I have missed, but I cannot see a way to generate the new Open Test Reporting format supported by JUnit 5 and 6 when running tests using java_test_suite.

If there is a way to do this then I would love to know, but if there is not, then perhaps a configuration parameter to specify what style of report should be generated would be useful.  

Obviously it would default to the current behavior, and different JUnit versions would support different report types.  Perhaps the parameter should take a list of values in case multiple reports were wanted, but they would obviously need a naming convention (to avoid all the files being called test.xml).

Please do let me know if this is a silly idea...

PR #432, opened 2026-02-20
Add the ability to generate additional JUnit 5 test reports

amilbourne:
This PR stems from issue #410

This adds 2 extra parameters to the java_junit5_test target.  They are 'report_generator' and 'report_generator_deps'.  This allows the 2 standard JUnit5 reports (legacy & open) to be generated as well as user defined reports.

The additional tests demonstrate how to use the new parameters, which should generate additional reports, rather than changing the existing report (test.xml).

Because open test reporting was only added in JUnit 5.9, I upgraded the JUnit dependencies from 5.8 to 5.9.  I think the target should still work fine with JUnit <5.9 as I haven't directly depended on the open reporting generator.  The target will obviously fail if you specify open reporting when running against an earlier version of JUnit 5.  I upgraded in order to test open reporting, but if the upgrade is a problem then I could remove the test and all would be fine without the upgrade.  I could also split the upgrade into a separate commit or PR if that makes it better.


tjoneslo:
My concern with this is with the output files generated. 

In the default case, when no report engine is specified, this generates a `test<something>.xml` containing the test results. This outputs into the `bazel-testlog` directory as a `test.log` and a `test.xml` file. I'm not sure what the report generators output and I want to make sure the default files are either not generated, or will not be overwritten or interleaved (with potential breakages). 

Otherwise I like this idea and can see it could be very useful.

amilbourne:
Sorry for my slow response.  I did originally try writing some tests to check that the files were generated correctly but this was quite gnarly code as it involved calling a build from a test.  I could put this back and you can see if you think it is OK.  I will update this PR.

amilbourne:
@tjoneslo I have added improved tests which check the content of the generated test reports, with and without the new parameters.  The tests are not simple, and could perhaps be improved, but I hope they are not too bad.

PR #434, opened 2026-02-24
[gazelle] Detect fully qualified class names in kotlin code

shs96c:


stevebarrau:
Is this a test only change?

shs96c:
Apparently. Don't think we need this. Closing

PR #435, opened 2026-02-25
[gazelle]: Fix missed Java deps from static calls on bare class names

shs96c:
Gazelle was missing valid Java dependencies when static methods were called on bare class identifiers, especially in same-package references and imported acronym-style types like `UUID`. This caused underreported `usedTypes`, which can lead to incorrect dependency inference.

This change closes that gap so dependency detection better matches real Java usage patterns and avoids silently dropping required deps in common static-call code paths.


shs96c:
No longer needed

PR #438, opened 2026-02-26
Reduce number of tests per PR

shs96c:
While it's easy to just run everything, all the time, it's wasteful and since the Windows tests are flaky it means we get flaky test runs. However, we do want to ensure that the embedded parser works as expected, so we can't _not_ run tests.

This attempts to strike a balance: running a subset of the tests so we can have confidence, but not too many.


shs96c:
Apparently not needed. Closing

PR #442, opened 2026-03-09
Fix checkstyle binary transition

Direnol:
This PR fixes an issue when a client build with remote in case an exec platform is different (aarch64 -> amd64 -> aarch64)

We need to bring a correct configuration for checkstyle binary that will be executed on a **target** platform

Without this change, you will get:

```
...-checkstyleexec.runfiles/rules_java++toolchains+remotejdk25_linux/bin/java: cannot execute binary file
```


shs96c (approved):
LGTM. I think we'll likely need the same fix for PMD and Spotbugs.

PR #454, opened 2026-05-13
Fix log message error levels.

mkosiba:
The message about `java_export` targets doesn't cause gazelle to fail, so it should be a warning.
The message about not being able to find a package will cause gazelle to fail, so it should be an error.


sitaktif (approved):
LGTM - fyi there's a small typo in the commit message

PR #455, opened 2026-05-27
[javaparser] Close StandardJavaFileManager to stop leaking file descriptors

shs96c:
## Summary
- `ClasspathParser.parseClasses(Path, List<String>)` constructs a `StandardJavaFileManager` and never closes it. Each call leaks both the native file-manager handle and the FDs of the `JavaFileObject`s it produces.
- Long-lived JVMs (e.g. the persistent javaparser gRPC server) can exhaust the per-process FD limit after enough invocations.
- Fix: wrap the manager in try-with-resources.


stevebarrau (approved):
Thanks @shs96c !

PR #456, opened 2026-05-29
xml reporting: group Vintage XML suites by class

strdn:
Use the nearest suite parent when Vintage descriptors have no class segment, so test cases are not reported as separate suites.


shs96c (approved):
LGTM. Thank you!

PR #457, opened 2026-06-02
Expose KtParser for external use

shs96c:


shs96c:
We probably don't want to make this public.

PR #460, opened 2026-06-29
[gazelle] Resolve classes from Maven classifier jars

shs96c:
The resolver seeded the index only by the plain `group:artifact` coordinates listed in the lock file, so classes that exist only in a classifier jar (e.g. `test-fixtures`, `unshaded`) were never resolved. Seed those from the classifier-keyed index entries (`group:artifact:extension:classifier`) so they resolve to `@maven//:group_artifact_classifier`.


java/gazelle/private/maven/coordinate.go
```diff
@@ -43,3 +43,25 @@ func (c *coordinate) ArtifactString() string {
 	}
 	return strings.Join(parts, ":")
 }
+
+// parseIndexKey parses a maven_index.json key: a versionless Maven coordinate in
+// rules_jvm_external's asKey form, group:artifact[:extension[:classifier]]. Plain,
+// packaging-qualified, and classifier-qualified keys are all handled here, so every
+// index entry maps to a dependency label the same way via ArtifactString.
+func parseIndexKey(key string) (*coordinate, bool) {
```
stevebarrau: Can we return an error here instead of a bool ?

java/gazelle/private/maven/resolver.go
```diff
@@ -134,36 +136,69 @@ func NewResolver(opts ...ResolverOption) (Resolver, error) {
...
+	}
+
+	// Seed package and class data from the index. Every index key is an artifact's
+	// versionless coordinate (group:artifact[:extension[:classifier]]); plain,
+	// packaging-qualified, and classifier-qualified keys are all seeded the same
+	// way, so a class living only in a classifier jar (e.g. test-fixtures) resolves
+	// just like any other.
+	if index != nil {
+		for _, key := range indexKeys(index) {
+			if coords, ok := parseIndexKey(key); ok {
```
stevebarrau: In what context would this fail? Is this something we want to log to the user?

java/gazelle/private/maven/resolver_test.go
```diff
@@ -49,3 +49,45 @@ func assertResolves(t *testing.T, r Resolver, excludePackages map[string]struct{
...
+func TestResolverClassifier(t *testing.T) {
+	r, err := NewResolver(
+		WithInstallFile("testdata/classifier_maven_install.json"),
+		WithIndexFile("testdata/classifier_maven_index.json"),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	none := map[string]struct{}{}
```
stevebarrau: Nit.
```suggestion
	none := make(map[string]struct{})
```

java/gazelle/private/maven/resolver.go
```diff
@@ -134,36 +136,69 @@ func NewResolver(opts ...ResolverOption) (Resolver, error) {
...
+				r.seedIndexKey(index, key, coords.ArtifactString())
 			}
 		}
 	}
 
 	return &r, nil
 }
 
+// indexKeys returns the sorted union of the index's package and class keys.
+func indexKeys(index *IndexFile) []string {
```
stevebarrau: What does the sort convey here? a fully-qualified class present in both a plain and a classifier jar now resolves to whichever key sorts last — the classifier (the plain coordinate is a string prefix, so it sorts first and gets overwritten). Previously nondeterministic, so this is strictly better, but if there’s a “right” answer it’s arguably the plain jar.

java/gazelle/private/maven/coordinate.go
shs96c: Sure. Done.

java/gazelle/private/maven/resolver.go
shs96c: It really shouldn't fail, as it indicates we've got some wonky coordinates in the index file. I'll make this a fatal error.

java/gazelle/private/maven/resolver_test.go
shs96c: Done

java/gazelle/private/maven/resolver.go
shs96c: We can let the plain jar win. I'll add a test.

stevebarrau (approved):

PR #465, opened 2026-07-02
[gazelle] Kotlin import-resolution fixes

shs96c:
Two small, independent Kotlin import-resolution fixes.

- Deeply-nested member imports (e.g. `StringResources.a.b.c`) were mapped only to their immediate parent, leaving a phantom package that never resolves. Walk up to the outermost class-like segment so the longest class-rooted prefix becomes the resolvable type.
- `kotlin.test` is in the `kotlin` namespace but ships as a separate artifact (`org.jetbrains.kotlin:kotlin-test`), not `kotlin-stdlib`. Exclude it from the stdlib check so it resolves to a Maven dependency.

### DIFF
--- a/java/gazelle/private/kotlin/kotlin.go
+++ b/java/gazelle/private/kotlin/kotlin.go
@@ -6,6 +6,14 @@ import (
 
 var kotlinStdLibPrefix = types.NewPackageName("kotlin")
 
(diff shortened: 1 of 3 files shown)


stevebarrau:
Please add to the IsStdlib tests.

stevebarrau (approved):

PR #466, opened 2026-07-02
[gazelle] Add java_testonly directive for test-support source sets

shs96c:
Gradle `testFixtures` and similar source sets are production code that exists only to support tests, but `IsTestPackage` cannot auto-detect them. This adds a `# gazelle:java_testonly` directive (inherited by child packages) that marks every generated production library `testonly`, keeping it out of production targets' provider set so a package it shares with `main` resolves unambiguously to the main library.

Visibility is now set regardless of `testonly` (previously only the non-testonly branch set it): a testonly library still needs to be visible to its testonly consumers in other packages. This also gives generated `java_test_suite` helper libraries visibility.


java/gazelle/generate.go
```diff
@@ -365,7 +365,7 @@ func (l javaLang) GenerateRules(args language.GenerateArgs) language.GenerateRes
 			}
 		}
 
-		l.generateJavaLibrary(args.File, args.Rel, cfg.MapLibraryName(filepath.Base(args.Rel)), productionJavaFiles.SortedSlice(), resourcesDirectRef, resourcesRuntimeDep, allPackageNames, nonLocalProductionJavaImports, nonLocalProductionJavaImportedClasses, nonLocalJavaExports, nonLocalJavaExportedClasses, nonLocalJavaExternalExportedClasses, annotationProcessorClasses, false, javaLibraryKind, &res, cfg)
+		l.generateJavaLibrary(args.File, args.Rel, cfg.MapLibraryName(filepath.Base(args.Rel)), productionJavaFiles.SortedSlice(), resourcesDirectRef, resourcesRuntimeDep, allPackageNames, nonLocalProductionJavaImports, nonLocalProductionJavaImportedClasses, nonLocalJavaExports, nonLocalJavaExportedClasses, nonLocalJavaExternalExportedClasses, annotationProcessorClasses, cfg.TestOnly(), javaLibraryKind, &res, cfg)
```
stevebarrau, line 368: At some point we are going to have to move all those into a struct 😅

stevebarrau (approved):

PR #467, opened 2026-07-02
[gazelle] Resolve test deps on test-suite helpers and testonly split packages

shs96c:
Stacked on #466 (`java_testonly` directive); review/merge that first. Its commit shows in this diff until it lands.

When a test imports a class its resolved production dependency does not declare — a class from another package's `java_test_suite` helper library, or from a testonly library sharing that package (e.g. a `testFixtures` source set) — the package resolves unambiguously to the production target and the class is missed. The stock test-suite resolution only fires when the package has *no* production provider, so this split main/test-suite case falls through.

In the unambiguous class-resolution path this now also consults the in-repo class index for testonly providers and the lone `java_test_suite`'s `<suite>-test-lib`, depending on the helper lib only when it actually declares the class (a class the production provider doesn't declare may be a `main` top-level function, not a test helper). Suite helper classes are cached so that check is exact.


java/gazelle/resolve.go
```diff
@@ -621,6 +639,33 @@ func (jr *Resolver) ruleDeclaresClass(lbl label.Label, className types.ClassName
...
+// resolveTestSuiteHelperClass returns the "<suite>-test-lib" helper library of the lone
+// java_test_suite that provides imp AND declares className, or NoLabel otherwise.
+// java_test_suite moves its non-test sources into a helper library named <suite>-test-lib;
+// a test in another package that imports one of those helpers must depend on it. This mirrors
+// resolveSinglePackageWithAmbiguity's test-suite branch, which only fires when imp has no
+// production provider -- here we cover the case where it has one (a split main/test-suite
+// package), so the package itself resolves to production and only the helper class is missing.
+// The className check guards against false positives: a class the production provider does not
+// declare may be a main top-level function (not registered as a class), which the helper lib
+// does not declare either and must not pull in.
```
stevebarrau: Please sort out the doc here, IIUC the resolveClassFromCrossResolver doc should move down and be attached to the resolveClassFromCrossResolver function below.

shs96c: Done

stevebarrau:
Please provide a test for this.

stevebarrau (approved):

PR #469, opened 2026-07-02
[gazelle] Filter own-module exports at resolve time

shs96c:
Under `# gazelle:java_module_granularity module` (and any other aggregate-at-root granularity), sub-package exports can name classes in other packages of the same module. This happens most commonly as a Kotlin extension function whose receiver or return type lives in a sibling sub-package. `addFilteringOutOwnPackage` only knows the per-file `ownPackage`, not the module-wide set, so those cross-package refs land in `nonLocalJavaExports` / `nonLocalJavaExternalExportedClasses` and, at resolve time, leak into the `exports` attr as own-module packages
the rule itself provides.

Extract the imports-side filter into `filterPackagesInModule` / `filterClassesInModule` helpers and apply them to the exports side too, so `allPackageNames` (the module-wide provider set) strips own-module refs from both sides symmetrically.


java/gazelle/generate.go
```diff
@@ -632,6 +625,37 @@ func addNonLocalImportsAndExports(toImports *sorted_set.SortedSet[types.PackageN
...
+	moduleSlice := modulePackages.SortedSlice()
+	return classes.Filter(func(c types.ClassName) bool {
+		for _, m := range moduleSlice {
+			if c.PackageName().Name == m.Name {
+				return false
+			}
+		}
+		return true
+	})
+}
```
stevebarrau: Both helpers reimplement set membership as an `O(n·m)` nested linear scan when `SortedSet.Contains` already exists.

shs96c: Done

stevebarrau:
Can we add a test for this?

shs96c:
Tests added.

stevebarrau (approved):

PR #472, opened 2026-07-08
Add opt-in invocation-level sharding for parameterized tests

### LINKED ISSUE #279: Sharding does not work with ParametrizedTest in java_junit5_test
All tests generated by [@ParametrizedTest](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests) are executed in the single shard.

A simple example is available in the [repo](https://github.com/al-babych-fivetran/shard-example).
It has a parametrized test with 20 values:
```
@ParameterizedTest
  @ValueSource(ints = {1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20})
  public void passes(int value) throws InterruptedException {
    //...
    System.out.println("Test: " + value + ", Shard: " + shard);
   // ...
  }
```

The `java_junit5_test` has `shard_count` equal to 10.
All tests are executed in a single shard:
```
Test: 1, Shard: 8
Test: 19, Shard: 8
Test: 4, Shard: 8
Test: 5, Shard: 8
Test: 2, Shard: 8
Test: 3, Shard: 8
Test: 8, Shard: 8
Test: 9, Shard: 8
Test: 6, Shard: 8
Test: 7, Shard: 8
Test: 10, Shard: 8
Test: 20, Shard: 8
Test: 14, Shard: 8
Test: 13, Shard: 8
Test: 12, Shard: 8
Test: 11, Shard: 8
Test: 18, Shard: 8
Test: 17, Shard: 8
Test: 16, Shard: 8
Test: 15, Shard: 8
```

I see [why](https://github.com/bazel-contrib/rules_jvm/blob/main/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestSharding.java#L26) this happened but is there any workaround for this?

helenyugithub:
Fixes #279.

## Problem

The JUnit5 runner implements Bazel test sharding as a `PostDiscoveryFilter`, which assigns each test descriptor to a shard by hashing its unique ID. Post-discovery filters run before test templates (`@ParameterizedTest`, `@RepeatedTest`) have generated their invocations — the invocations simply do not exist at discovery time — so a template can only be filtered as a single unit, and all of its invocations land on the same shard.

This pins the full cost of a large parameterized test onto one shard. As a data point from a large monorepo: one sub-shard of an acceptance-test target consistently ran ~760s while its sibling shards ran ~90s, entirely because of this behaviour. The available workaround (manually splitting a parameterized method into several wrapper methods, each running a partition of the arguments) works but is pure boilerplate.

## Approach

Since invocation-level filtering cannot happen at discovery time, this PR splits the decision across discovery and execution, behind a new opt-in annotation:

- **`@ShardTemplateInvocations`** (`METHOD`/`TYPE`): opts a test template — or every template in a class — into invocation-level sharding. It is meta-annotated with `@ExtendWith(TemplateInvocationShardingCondition.class)`, so applying it registers the condition; no global extension registration is involved, and behaviour is unchanged for everyone who does not use the annotation.
- **Shard filter** (`TestSharding`): descriptors for annotated test templates (detected via `AnnotationSupport.findAnnotation` on the `MethodSource` method, its class, superclasses, and enclosing classes) are now included on *every* shard instead of being hash-assigned to one. Any reflective failure falls back to the existing single-unit behaviour.
- **`TemplateInvocationShardingCondition`** (`ExecutionCondition`): at execution time, enables each `test-template-invocation` on exactly one shard, assigning invocation `N` (1-based) round-robin to shard `(N - 1) % TEST_TOTAL_SHARDS`. Round-robin is deterministic and keeps per-shard invocation counts balanced. Containers pass through, and the condition is a no-op when sharding is not enabled. Invocations belonging to other shards show up as skipped in that shard's XML output.

Caveats, documented on the annotation and in the runner README:

- The set and order of invocations must be stable across shards (deterministic argument providers), otherwise invocations may run on multiple shards or none.
- `@TestFactory` is not supported: dynamic tests cannot be conditionally disabled, so the filter deliberately applies the include-on-every-shard behaviour only to `test-template` descriptors and factories continue to be sharded as a single unit (including a factory everywhere would duplicate its tests on every shard).

## Testing

- New `invocation-sharding-test` (`shard_count = 3`), mirroring the existing `sharding-test`: runs the JUnit Platform launcher in-process with the shard filter installed against a new sample class (kept in a separate `java_library` so the existing `sharding-test` golden assertions are untouched) containing an annotated `@ParameterizedTest` with 7 invocations, an annotated `@RepeatedTest(5)`, and a non-annotated `@ParameterizedTest`. Each shard asserts the exact set of started and skipped descriptors: annotated templates start on every shard with exactly the round-robin slice of invocations started and the rest skipped, while the non-annotated template retains the existing behaviour (all invocations on a single hash-assigned shard).
- New `invocation-sharding-unsharded-test`: runs the same sample class without sharding and asserts every invocation runs and nothing is skipped.
- The existing `sharding-test` golden per-shard sets pass unchanged.

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/README.md
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/README.md
@@ -18,6 +18,50 @@ tag. This can contain any number of `testsuite` tags. Each of these
 `testsuite` tags may only contain any number of nested `testcase` tags. The 
 model does not allow nesting of `testsuite` tags within `testsuite` tags. 
 
+## Test sharding
+
+The runner supports [Bazel test sharding][sharding] via a `PostDiscoveryFilter`
+which assigns each test to a shard based on a hash of its unique ID. Because
+post-discovery filters run before test templates (such as `@ParameterizedTest`
+or `@RepeatedTest` methods) have generated their invocations, a test template
+is assigned to a shard as a single unit by default: all of its invocations run
+on the same shard.
+
+### Sharding the invocations of a test template
+
+If a single parameterized or repeated test dominates the runtime of a shard,
+you can opt in to distributing its invocations across shards by annotating the
+test method (or its class, to opt in every test template in the class) with
+`@ShardTemplateInvocations`:
+
+```java
+@ShardTemplateInvocations
+@ParameterizedTest
+@ValueSource(ints = {0, 1, 2, 3, 4, 5, 6})
+void myTest(int value) {
+  // ...
+}
+```
+
+The test template is then included on every shard, and each invocation is
+enabled on exactly one shard: invocation `N` (1-based) runs on shard
+`(N - 1) % TEST_TOTAL_SHARDS`, which is deterministic and keeps per-shard
+invocation counts balanced. Invocations belonging to other shards are reported
+as skipped in that shard's XML output.
+
+Note that:
+
+* The set and order of invocations must be stable across shards: every shard
+  must generate the same invocations in the same order, so argument providers
+  must be deterministic. If they are not, some invocations may run on more
+  than one shard or on no shard at all.
+* `@TestFactory` methods are not supported: dynamic tests cannot be
+  conditionally disabled, so test factories are always sharded as a single
+  unit.
+* The annotation has no effect when test sharding is not enabled.
+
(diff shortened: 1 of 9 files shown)


tjoneslo (approved):
This looks great. Look forward to seeing it in action.

PR #474, opened 2026-07-17
Time-bound and cache the hostname lookup in TestSuiteXmlRenderer

### LINKED ISSUE #473: JUnit5 runner hangs indefinitely in TestSuiteXmlRenderer.getHostname() when the host resolver is unresponsive
## Summary

`TestSuiteXmlRenderer.getHostname()` calls `InetAddress.getLocalHost()` while rendering each suite's XML report. On a machine whose DNS resolver becomes unresponsive (observed on ephemeral Linux CI runners whose hostname is not present in `/etc/hosts`, so the lookup always leaves the machine), that call blocks indefinitely — **after all tests have already passed** — and the test JVM never exits. In our CI this manifested as test actions hanging for hours.

A second, compounding issue: the hang occurs while `BazelJUnitOutputListener.output()` holds the listener lock. The shutdown hook needs the same lock to flush results, so `SIGTERM` cannot produce a clean exit either — the JVM lingers until `SIGKILL`.

## Environment

- contrib_rules_jvm 0.33.0, `java_test_suite` with `runner = "junit5"`
- OpenJDK 25.0.2 (GraalVM CE), Linux x86_64 CI runners
- Runner hostname absent from `/etc/hosts` → `getLocalHost()` performs a real resolver round-trip on every suite completion

## Thread dump (captured via SIGQUIT ~7.6 min into the hang; all tests had passed)

```
"main" #3 prio=5 cpu=1820.50ms elapsed=459.42s  RUNNABLE
    at java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method)
    at java.net.InetAddress$PlatformResolver.lookupByName(InetAddress.java:1134)
    at java.net.InetAddress.getAddressesFromNameService(InetAddress.java:1712)
    at java.net.InetAddress$NameServiceAddresses.get(InetAddress.java:1060)
    at java.net.InetAddress.getAllByName0(InetAddress.java:1702)
    at java.net.InetAddress.getLocalHost(InetAddress.java:1796)
    at com.github.bazel_contrib.contrib_rules_jvm.junit5.TestSuiteXmlRenderer.getHostname(TestSuiteXmlRenderer.java:88)
    at com.github.bazel_contrib.contrib_rules_jvm.junit5.TestSuiteXmlRenderer.toXml(TestSuiteXmlRenderer.java:32)
    at com.github.bazel_contrib.contrib_rules_jvm.junit5.BazelJUnitOutputListener.output(BazelJUnitOutputListener.java:219)
    - locked <0x0000000088ffe570> (a java.lang.Object)          // lock the shutdown hook also needs
    at com.github.bazel_contrib.contrib_rules_jvm.junit5.BazelJUnitOutputListener.outputIfTestRootIsComplete(BazelJUnitOutputListener.java:188)
    at com.github.bazel_contrib.contrib_rules_jvm.junit5.BazelJUnitOutputListener.executionFinished(BazelJUnitOutputListener.java:180)
    ... [JUnit Platform launcher plumbing elided] ...
    at com.github.bazel_contrib.contrib_rules_jvm.junit5.ActualRunner.run(ActualRunner.java:121)
    at com.github.bazel_contrib.contrib_rules_jvm.junit5.JUnit5Runner.main(JUnit5Runner.java:37)
```

(All other threads were idle JVM housekeeping; no application thread was blocked on anything else. `RUNNABLE` with 1.8s CPU over 459s elapsed = blocked in the native resolver call, not spinning.)

## Suggested fix

- Resolve the hostname **once** (static cache) instead of per suite render, and/or
- Time-bound the lookup (resolve on a separate thread with a small timeout) and fall back to `localhost` / `$HOSTNAME` env when it doesn't answer promptly. The hostname is cosmetic metadata in the XML — it should never be able to hang the run.

## Workaround

Pinning the runner hostname in `/etc/hosts` (`echo "127.0.0.1 $(hostname)" >> /etc/hosts`) makes the lookup resolve from files and fully mitigates the hang.

helenyugithub:
Fixes #473.

`TestSuiteXmlRenderer.getHostname()` called `InetAddress.getLocalHost()` on every suite render. That consults the system resolver, and on a host whose resolver is unresponsive (and whose hostname is absent from `/etc/hosts`) the native lookup blocks **indefinitely** — the existing `catch → "localhost"` never helps because the failure mode is a hang, not an exception. A test JVM whose tests had already passed could hang forever writing the XML report; and because the hang occurs while `BazelJUnitOutputListener.output()` holds its lock, the shutdown hook deadlocks on the same monitor, so `SIGTERM` can't produce a clean exit either. Full thread dump from a real occurrence (Linux CI, contrib_rules_jvm 0.33.0) is in #473.

This change:

- resolves the hostname **once per JVM** (static cache) instead of per suite render;
- performs the lookup on a **daemon thread bounded at 5 seconds** (`FutureTask.get`), so a never-returning native lookup can neither block the render nor keep the JVM alive at exit;
- falls back to the `HOSTNAME` environment variable, then `"localhost"`, keeping the attribute meaningful on machines where the env var is set but the resolver isn't answering.

Verified locally: `bazel build //java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5` and `bazel test //java/test/com/github/bazel_contrib/contrib_rules_jvm/junit5/...` — 92/92 pass, including the checkstyle/pmd/spotbugs targets.

### DIFF
--- a/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestSuiteXmlRenderer.java
+++ b/java/src/com/github/bazel_contrib/contrib_rules_jvm/junit5/TestSuiteXmlRenderer.java
@@ -11,11 +11,20 @@ import java.time.Instant;
 import java.time.format.DateTimeFormatter;
 import java.util.Collection;
 import java.util.Locale;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.TimeUnit;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 import org.junit.platform.launcher.TestPlan;
 
 class TestSuiteXmlRenderer {
+
+  // Resolved once and time-bounded: InetAddress.getLocalHost() consults the system resolver,
+  // and on a host with an unresponsive resolver (and a hostname absent from /etc/hosts) the
+  // native lookup can block indefinitely. This is called while rendering results of completed
+  // tests, so an unbounded lookup hangs the JVM after the tests have already passed. The
+  // hostname is cosmetic report metadata; it must never be able to hang the run.
+  private static final String HOSTNAME = resolveHostname();
 
   private final TestCaseXmlRenderer testRenderer;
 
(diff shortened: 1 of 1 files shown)


shs96c (approved):
LGTM

shs96c:
Thank you for your contribution!

PR #475, opened 2026-07-17
feat(gazelle): add java_parser directive for Turbine-based Java parsing

shs96c:
Adds a `# gazelle:java_parser` directive (values: `javac`, `turbine`; default: `javac`) that selects the Java parser used during the Generate phase. The directive is inherited by child packages, so a single entry in the root BUILD file applies everywhere.

`turbine` uses Google's [Turbine](https://github.com/google/turbine) header-only parser, which is ~50x faster than javac on large repos (measured: 510s → 9s on 111k files). An automatic per-file fallback to javac handles files with unresolved `.class` literals (split-package targets where a same-package type is referenced via `Foo.class` without an explicit import).

No behavioural changes when `java_parser` is not set.


shs96c:
This isn't as good as we'd hoped -- it misses way too many cases.

PR #484, opened 2026-09-08
Make `HOSTNAME` env var short-circuit lookup in `TestSuiteXmlRenderer`

### LINKED ISSUE #483: Difficult to work around 5s hang on unresponsive host resolver in JUnit5 runner
# Problem

When rendering a suite's XML report `TestSuiteXmlRenderer` calls `InetAddress.getLocalHost()` to discover the hostname. In some environments where the DNS resolver is unresponsive this might hang until the 5s timeout. There's a fallback to the environment variable `$HOSTNAME` so such environments can still be supported. However, setting `$HOSTNAME` does not avoid the 5s hang on each run in those environments.

# Background

https://github.com/bazel-contrib/rules_jvm/pull/474 introduced a 5s time-bound on the hostname lookup in `TestSuiteXmlRenderer` to address https://github.com/bazel-contrib/rules_jvm/issues/473. In addition, a fallback to `$HOSTNAME` added.

# Suggested Fix

Let `$HOSTNAME` take precedence over the resolver lookup so the hostname discovery is short-circuited before lookup when it is set.

In most cases when `$HOSTNAME` is set, we can expect it to be equivalent to the lookup. It's a cosmetic feature, so even if it does diverge it's very unlikely to be a breaking change.

NilsBarlaug:
Fixes https://github.com/bazel-contrib/rules_jvm/issues/483.

Currently, `TestSuiteXmlRenderer` tries to lookup the hostname by calling `InetAddress.getLocalHost().getHostName()`. Since this might hang indefinitely, we have a 5s timeout. If we fail to look it up, we fall back to `$HOSTNAME` (or `"localhost"`) to allow using a meaningful value in environments where the lookup isn't possible. However, there's no way to avoid the 5s hang when the lookup isn't possible.

This changes `TestSuiteXmlRenderer.resolveHostname()` to short-circuit if `$HOSTNAME` is set. That is, if `$HOSTNAME` is set we simply use that and don't try to lookup the hostname through `InetAddress.getLocalHost().getHostName()` at all.

In theory, this could change the resolved hostname in some environments where the lookup is available and `$HOSTNAME` is set to something different. However, we expect most environments that have `$HOSTNAME` set to have it set to the equivalent of the lookup - and even when that's not the case it's still a cosmetic feature very unlikely to be a breaking change.


shs96c (approved):
Thank you! Looks like this replaces #480 which was trying to solve the same problem.

PR #487, opened 2026-09-11
fix(gazelle): remove generated java targets when proto_library is removed

stevebarrau:
## Problem

When a `proto_library`'s `.proto` sources are deleted, Gazelle's built-in proto language deletes the now-empty `proto_library` rule. However, the Java Gazelle extension left the Java targets it had generated for that `proto_library` behind as dangling rules:

- `java_proto_library`
- `java_grpc_library`
- `java_library`

Running Gazelle after removing a proto therefore produced BUILD files that still referenced a `proto_library` target that no longer existed.

## Root cause

The proto language reports rules it wants deleted via `GenerateArgs.OtherEmpty` (rules that merge to empty are then removed by Gazelle). The Java extension's `generateProtoLibraries` only inspected `args.OtherGen`, so it never learned that a `proto_library` was being removed and never emitted the corresponding empty Java rules.

## Fix

Iterate over `args.OtherEmpty` in `generateProtoLibraries`, and for each emptied `proto_library` emit matching empty `java_proto_library`, `java_grpc_library`, and `java_library` rules into `res.Empty`. Gazelle only deletes a rule if it merges to empty against the existing BUILD file, so emitting an empty rule for a name that isn't present is a harmless no-op. The Java rule-name derivation is factored into a small shared `protoJavaRuleNames` helper used by both the generate and remove paths.

## Test

Adds a `proto_removal` generation test: a `BUILD.in` containing a `proto_library` plus its generated Java targets, but with no `.proto` file present on disk. The expected `BUILD.out` is empty, verifying that all of the proto and Java targets are removed. Confirmed the test fails without the fix and passes with it.


java/gazelle/testdata/proto_removal/maven_install.json
```diff
@@ -0,0 +1 @@
+{"version": "2"}
```
shs96c, line 1: Nothing to do here, but the current version is now 3

shs96c (approved):
LGTM

PR #489, opened 2026-09-15
feat: add per-test suite overrides to `java_test_suite`

shs96c:
There are occasions where it's helpful to override specific attributes for tests within a suite (eg. to set a tag such as `no-remote-exec`, or to set a shard count). Without an escape hatch, there's no easy way to do this.


java/private/create_jvm_test_suite.bzl
```diff
@@ -112,41 +136,30 @@ def create_jvm_test_suite(
         **library_attrs
     )
 
-    # Get any deps referenced in make vars
-    make_var_fields = (kwargs.get("jvm_flags", []) +
-                       kwargs.get("javacopts", []) +
-                       kwargs.get("args", []) +
-                       kwargs.get("env", {}).values())
-
-    # Handle select() in deps or make_var_fields - select is not iterable, so skip make_var extraction
```
tjoneslo, line 121: Can you move this comment to the `_make_var_deps()` method to explain why it's returning empty lists when the values are not lists.

tjoneslo (approved):
One requested update to keep a comment for the next person trying to understand what is going on. Otherwise LGTM

stevebarrau (approved):

